PR #53, opened 2019-02-28
readme improvement

paulbrie:


src/shared/utils/helpers.ts
```diff
@@ -19,3 +19,15 @@ export const stringToCamelCase = (str: string): string =>
   str.toLowerCase().replace(/[^a-zA-Z0-9]+(.)/g, (_, chr) => chr.toUpperCase())
 
 export const capitalize = (str: string): string => str[0].toUpperCase() + str.slice(1)
+
```
alexnm: we should rename `helpers.ts` to `string-utils.ts`, as all the functions perform string manipulation. I'll do that a bit later

src/component-generators/vue/vue-component.ts
```diff
@@ -44,22 +46,23 @@ const createVueGenerator = (
     const cssChunks = result.chunks.filter((chunk) => chunk.meta.fileId === 'vuecss')
     const htmlChunks = result.chunks.filter((chunk) => chunk.meta.fileId === 'vuehtml')
 
-    const jsCode = chunksLinker.link(jsChunks)
-    const cssCode = chunksLinker.link(cssChunks)
-    const htmlCode = chunksLinker.link(htmlChunks)
+    const jsCode = removeLastEmptyLine(chunksLinker.link(jsChunks))
```
alexnm: does it make sense to move the `removeLastEmptyLine` call inside the `link` function? does it affect the React component generation?

src/component-generators/vue/vue-component.ts
```diff
@@ -44,22 +46,23 @@ const createVueGenerator = (
...
-      code: `
-<template>
-${htmlCode}</template>
+      code: `<template>
+${addSpacesToEachLine(2, htmlCode)}
+</template>
 
 <script>
-${jsCode}</script>
+${addSpacesToEachLine(2, jsCode)}
```
alexnm: as far as I know, Vue developers prefer to not have the script and styles indented, only the template.

A few examples (didn't find one with both styles and script):
https://vuejs.org/v2/style-guide/#Element-selectors-with-scoped-use-with-caution
https://github.com/gothinkster/vue-realworld-example-app/blob/master/src/components/TagList.vue

src/component-generators/vue/vue-component.ts
paulbrie: I don't know yet and didn't want to generate collateral effects.

src/shared/utils/helpers.ts
paulbrie: ok, sounds good

examples/project-exporters/vue-basic/index.ts
```diff
@@ -34,7 +35,7 @@ const writeToDisk = async (
 //   console.log(JSON.stringify(result, null, 2))
 // }
 
-console.log(UIDLValidators.validateProject(projectJson))
```
alexnm: this is just for testing purposes so it can remain as a console.log to see the results of the validation when we run the generators (we can add a ignore comment)

PR #68, opened 2019-03-05
fix for unformatted css in generated html and js files

mihaiserban:
postcss prettier plugin added to html-to-string and js-ast-to-code

formats css in js, as well as css in generated html files


mihaiserban:
any way to expose the travis ci build error?

alexnm:
actually I'm just doing the same change as I have the same fail on another PR, so you can just pull master into your branch

PR #77, opened 2019-03-15
Feature/changelog

JayaKrishnaNamburu:
Fixes #54 

Hi @alexnm I gave a try for both [semantic-release](https://github.com/semantic-release/semantic-release) and [bump](https://github.com/fabiospampinato/bump) well semantic-release looked very matured but after trying out to implement it on an example repo, I realized ``bump`` is something that we need here. Here are the observations for both

Pros for ``semantic-release``
- Can be integrated with Travis so releases happen automatically depending on the commit messages.
- A release is published in both GitHub and npm as well.

Cons for this case as the whole eco-system is in beta/alpha

- By default ``semantic-release`` starts its versioning with v1.0
- Maintaining pre-release, beta is a little tricky. 
- The configuration is a little [tricky](https://github.com/semantic-release/npm/issues/151) 

Why Bump?

- Generates changelog and updates it whenever we bump a ``major/minor/path`` is added.
- Releasing a version is totally controlled by the developer using ``bump`` command (That is where semantic-release helps to reduce human errors but it helps in production releases the most).
- Pushes to GitHub and drafts a release with selected fields and updates the changelog using ``bump.json`` file. 

~~~
bump changelog // generates changelog
bump // gives a list of options to raise the bump
bump release // releases/drafts to GitHub
~~~

Screenshot 2019-03-15 at 11 46 29 AM
Screenshot 2019-03-15 at 11 46 48 AM

PR #82, opened 2019-03-20
refactor(schemas): Updated schemas with enum values

### LINKED ISSUE #18: Use enum values in JSON Schema
Some of the fields that we are restricting as strings can be expressed by enums for better suggestions and better validation rules:
- globals.assets.type (link, script, style, etc.)
- globals.manifest - multiple fields (check specification)
- dependency.type (local, package, library)
- propDefinitions.type
- stateDefinitions.type

JayaKrishnaNamburu:
fixes #18

### DIFF
--- a/src/uidl-definitions/schemas/component.json
+++ b/src/uidl-definitions/schemas/component.json
@@ -189,7 +189,7 @@
           "additionalProperties": false,
           "required": ["type"],
           "properties": {
-            "type": {"type": "string"},
+            "type": {"type": "string", "enum": ["string", "boolean", "number", "array", "func", "object", "children"]},
             "defaultValue": {},
             "meta": {"type": "object"}
           }
@@ -204,7 +204,7 @@
           "additionalProperties": false,
           "required": ["type", "defaultValue"],
           "properties": {
-            "type": {"type":"string", "enum": ["string", "boolean", "number", "object"]},
+            "type": {"type":"string", "enum": ["string", "boolean", "number", "object", "func", "array"]},
             "values": {
               "type": "array",
               "items": {
--- a/src/uidl-definitions/schemas/component.json
+++ b/src/uidl-definitions/schemas/component.json
@@ -164,7 +164,11 @@
       "additionalProperties": false,
       "required": ["type"],
       "properties": {
-        "type": {"type": "string", "examples": ["package", "local", "library"]},
+        "type": {
+          "type": "string",
+          "enum": ["package", "local", "library"],
+          "examples": ["package", "local", "library"]
+        },
         "path": {"type": "string"},
         "version": {"type": "string", "default": "1.0.0"},
         "meta": {
@@ -200,7 +204,7 @@
           "additionalProperties": false,
           "required": ["type", "defaultValue"],
           "properties": {
-            "type": {"type": "string"},
+            "type": {"type":"string", "enum": ["string", "boolean", "number", "object"]},
             "values": {
               "type": "array",
               "items": {
--- a/src/uidl-definitions/types/index.ts
+++ b/src/uidl-definitions/types/index.ts
@@ -33,13 +33,13 @@ export interface ComponentUIDL {
 
 export interface PropDefinition {
   type: string
-  defaultValue?: string | number | boolean | any[]
+  defaultValue?: string | number | boolean | any[] | object | (() => void)
   meta?: Record<string, any>
 }
 
 export interface StateDefinition {
   type: string
-  defaultValue: string | number | boolean | any[]
+  defaultValue: string | number | boolean | any[] | object | (() => void)
   values?: Array<{
     value: string | number | boolean
     meta?: {
--- a/src/uidl-definitions/types/index.ts
+++ b/src/uidl-definitions/types/index.ts
@@ -15,8 +15,16 @@ export interface ProjectUIDL {
   components?: Record<string, ComponentUIDL>
 }
 
+enum AssetType {
+  style = 'style',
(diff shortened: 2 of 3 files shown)


JayaKrishnaNamburu:
Hi, @alexnm I have updated the schema attributes using enum, please let me know if any further changes/corrections needed.

JayaKrishnaNamburu:
HI @alexnm i am a little new to typescript so for assigning ``enum`` for ``global.asset.type`` there are two ways, 

First either define an enum and cross check with that enum in all test cases like
~~~
export enum AssetType {
  "style",
}
if (asset.type === AssetType.style)
~~~
or declaring the ``asset.type`` in the following way
~~~
type: "style" | "script"
~~~

Please let me know which helps our use case here 😅

src/uidl-definitions/schemas/component.json
```diff
@@ -200,7 +204,7 @@
           "additionalProperties": false,
           "required": ["type", "defaultValue"],
           "properties": {
-            "type": {"type": "string"},
+            "type": {"type":"string", "enum": ["string", "boolean", "number", "object"]},
```
alexnm: same thing is needed for the type inside `propDefinition`
and you should also include `array` and `function`

alexnm:
@JayaKrishnaNamburu actually I gave it a bit more thought and it makes more sense to use the enums like you proposed for the asset type. the advantage with that is that we won't be writing strings in the code anymore if we have the enums. do you agree?

src/uidl-definitions/types/index.ts
```diff
@@ -120,6 +128,7 @@ export interface WebManifest {
   orientation?: string
   scope?: string
   theme_color?: string
+  screenshots?: Array<{ src: string; sizes: string; type: string }>
```
alexnm: What does this represent?

src/uidl-definitions/types/index.ts
```diff
@@ -15,8 +15,16 @@ export interface ProjectUIDL {
   components?: Record<string, ComponentUIDL>
 }
 
+enum AssetType {
```
alexnm: these changes should also be reflected in the schema, not only in the TS interfaces

JayaKrishnaNamburu:
> @JayaKrishnaNamburu actually I gave it a bit more thought and it makes more sense to use the enums like you proposed for the asset type. the advantage with that is that we won't be writing strings in the code anymore if we have the enums. do you agree?

Yes, I took agree with the point of not using strings anywhere in the codebase but was not sure with the scope of that implementation other than that use case. Personally I like to keep it this way but we need to import and use that ``enum`` when we try to use it In any other place in the codebase.

Will work on the 3 comments, thanks for the review 😄

src/uidl-definitions/types/index.ts
JayaKrishnaNamburu: It just holds the screenshots of the App but now i don't think it is not used much after [referring](https://www.w3.org/TR/appmanifest/#screenshots-member) to few examples, will remove that :D

src/uidl-definitions/types/index.ts
JayaKrishnaNamburu: Hey @alexnm i just pushed by making all the changes, please let me know if we go with this style of declaration in ts files so i will go ahead and update then and update all the files in codebase where we are using them.
~~~
export enum AssetType {
  "style",
}

alexnm: @JayaKrishnaNamburu looks great, please go ahead and add the TS enums. you can also check in the codebase where the strings are used and replace them. curious if it has any impact in running the tests, but all the generators should behave the same after this change.

JayaKrishnaNamburu: ok, will work changing them 👍

JayaKrishnaNamburu: @alexnm i don't think we can go ahead with this approach, here is the challenge in changing it into enum in ts files.

All Enum's are converted into objects in return so we will be able to using as ``AssetType.style`` in the code base but which helps in eliminating strings, but when we can't using ``AssetType.style`` in our project UIDL schema right, so the generators are breaking 😅  I need your inputs 😅

JayaKrishnaNamburu: declaring them in union style all the builds and projects are being generated without breaking

alexnm: please go with the union types for now, we'll figure out later if it makes sense to use enums or not. the @babel/types library we are using for manipulating ASTs is also using a similar approach as far as I can see

JayaKrishnaNamburu:
Hi @alexnm i just updated the PR, i just updated only the ``defaultValue`` for the types, because if i try to change the type here is the challenge i am facing currently.
In plugins like ``teleport-react-base-component` we have checks like this ``type: stateDefinitions[stateKey].type`` which is declared to check the type that is passed which is ``string`` so the ``generators`` and ``tests`` are breaking 😅  Please suggest your views on this.

alexnm:
@JayaKrishnaNamburu let's leave it like that for now, I'll merge this , I can take a look at the rest of the types when I'm done with #83, which involves some changes into props and state anyway

alexnm:
please take a look at #77 if you have time so we can also close that today

JayaKrishnaNamburu:
> please take a look at #77 if you have time so we can also close that today

Hey, @alexnm you want me to try out Lerna or any other alternatives or just make a proper ``bump.json`` so we can merge into master?

alexnm:
no, just check the comments left by @Utwo

Issue #55, opened 2019-03-01
 Publish declarations files under the @types organization on npm


mihaisampaleanu:
Publish teleport-code-generators declarations files under @types/teleport-code-generators

More info: https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html

alexnm:
one thing that is not clear to me, is it enough to specifiy the types path in `package.json` like in the doc and bundle them with the npm package? or you also should push them to the @types org on npm?

paulbrie:
As far as I know, it is not required if we already have them in our `package.json`. However, that forces the user to download everything even if the only need is just to type. We should check for best practices.

vladnicula:
<img width="1088" alt="Screenshot 2019-03-22 at 10 00 43" src="https://user-images.githubusercontent.com/126038/54808699-ad1b7880-4c89-11e9-8489-80d0888138d9.png">
It seems they make it very clear that the "best practice" for projects written in typescript, like ours, is to publish them with the package, not as a separate "@types/package-name" thing.

alexnm:
@mihaisampaleanu should we close this? we have the `types` linked in `package.json`

vladnicula:
@mihaisampaleanu @alexnm we don't have to close it. Someone needs to move the interfaces into a types declaration file, instead of exporting them and importing them where neeeded.

alexnm:
@vladnicula would you like to take this?

vladnicula:
on it.

PR #92, opened 2019-03-25
feat(Move type declarations to types file): The move to a central typ…

vladnicula:
…es file (.d.ts)

Required removing the babel types import in the declaration file as well as
removing many import statements.

re #55


vladnicula:
@alexnm this change will require playground changes as the exported names are no longer accessible from `GeneratorTypes`. 

There was also a issue with `import { Node } from @babel/types` which could not tree shake and was seeing the definition file as a normal file. I had to remove that. 

The question about `ReactFlavors` can be clarified by https://www.typescriptlang.org/docs/handbook/enums.html 

Screenshot 2019-03-25 at 17 20 52

Issue #99, opened 2019-03-26
Vue generator breaks if schema's name has '-' in declaration


JayaKrishnaNamburu:
So, here how to reproduce the scenario, I am sharing the schema sample too. In the example below the event's name is ``input-box`` so when the generator runs it's breaking and when I remove the ``-`` from the name the component is generated successfully.

Possible Solutions
1. Configure the ``schema`` not to accept ``-`` in the name if it has an event. (which is a little tricky and I am not sure how we can achieve).
2. Change the logic that is being currently used for generating Event Handler names.

```
           {
                "type": "text",
                "name": "input-box",
                "events": {
                    "onClick": [
                        {
                            "type": "stateChange",
                            "modifies": "toggleComponent",
                            "newState": true
                        }
                    ]
                }
            }
```

~~~
{ SyntaxError: Unexpected token, expected "," (11:16)
   9 | 
  10 |   methods: {
> 11 |     handleInput-boxOnClick() {
     |                ^
  12 |       this.toggleComponent = true;
  13 |     }
  14 | 
    at t (/Users/a1290641/Documents/personal/teleport-code-generators/node_modules/prettier/parser-babylon.js:1:264)
    at Object.parse (/Users/a1290641/Documents/personal/teleport-code-generators/node_modules/prettier/parser-babylon.js:1:208428
)
    at Object.parse$2 [as parse] (/Users/a1290641/Documents/personal/teleport-code-generators/node_modules/prettier/standalone.js:12953:19)
    at coreFormat (/Users/a1290641/Documents/personal/teleport-code-generators/node_modules/prettier/standalone.js:16180:23)
    at format (/Users/a1290641/Documents/personal/teleport-code-generators/node_modules/prettier/standalone.js:16441:73)
    at formatWithCursor (/Users/a1290641/Documents/personal/teleport-code-generators/node_modules/prettier/standalone.js:16457:12)
    at /Users/a1290641/Documents/personal/teleport-code-generators/node_modules/prettier/standalone.js:30711:15
    at Object.format (/Users/a1290641/Documents/personal/teleport-code-generators/node_modules/prettier/standalone.js:30719:12)
    at Object.exports.generator [as js] (/Users/a1290641/Documents/personal/teleport-code-generators/src/core/builder/generators/js-ast-to-code.ts:16:21)
    at Builder.generateByType (/Users/a1290641/Documents/personal/teleport-code-generators/src/core/builder/index.ts:82:33)
[...]

PR #102, opened 2019-03-27
WIP: Ternary Expressions or Vue Generators

### LINKED ISSUE #93: Optimize boolean states in <template>
Currently, a boolean state renders something like this
```
<span v-if="isExpanded">Collapse</span>
<span v-if="!isExpanded">Expand</span>
```

In terms of good practices for the vue generators, we might render boolean states correlated with `v-if` / `v-else`.

This is similar to #14, which we are doing for the React generators

JayaKrishnaNamburu:
fixes #93 #99

### DIFF
--- a/src/plugins/teleport-plugin-react-base-component/utils.ts
+++ b/src/plugins/teleport-plugin-react-base-component/utils.ts
@@ -9,6 +9,7 @@ import {
   addDynamicAttributeOnTag,
   generateASTDefinitionForJSXTag,
   createConditionalJSXExpression,
+  createTernaryOperation,
 } from '../../shared/utils/ast-jsx-utils'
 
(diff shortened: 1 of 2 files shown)


src/plugins/teleport-plugin-vue-base-component/utils.ts
```diff
@@ -70,23 +71,40 @@ export const generateVueNodesTree = (
         return
       }
 
-      if (child.type === 'state') {
+      if (child.type === 'state' && child.states) {
```
JayaKrishnaNamburu: Added this check to save ``0.33`` sec when there are no states declared, not sure if it is actually needed 😅

vladnicula: that's very specific :D

PR #104, opened 2019-03-27
feat(uidl validator): add validation layer for component generators

### LINKED ISSUE #2: Component generators should validate input UIDL
First step in the generators, before any node resolve, should be a call to the UIDL validator

Validator functions are found in: `/src/uid-definitions/validators`

Concerns:
- validation is done using an external JSON schema validation library (ajv) which may be slightly big in terms of kb and it would not be tree-shaked if used on all component generator paths
- validation should be opt-out, via a configuration in the generator

anamariaoros:
resolves #2


vladnicula:
@alexnm any ideas why these are failing?

PR #115, opened 2019-03-28
Refactor/#25 project generators

ovidiuionut94:


src/typings/generators.d.ts
```diff
@@ -232,6 +232,45 @@ interface GeneratedFile {
...
+  componentUIDL: ComponentUIDL
+  componentOptions: {
+    assetsPrefix: string
+    localDependenciesPrefix?: string
+  }
+  componentExtension?: string
+}
+
+interface FolderStructureParams {
+  distFolderName: string
```
alexnm: this interface looks like it doesn't scale well, can we think of different strategies for sending all the different files for different folders? Maybe with some sort of folder identifer?

src/shared/utils/project-utils.ts
```diff
@@ -161,3 +193,136 @@ export const createPackageJSON = (
...
+    },
+  }
+}
+
+export const joinComponentFiles = (
+  componentFiles: GeneratedProjectData[]
+): GeneratedProjectData => {
+  let dependencies = {}
+
+  const files = componentFiles.reduce((componentFile: GeneratedFile[], currentFile) => {
```
alexnm: you can reduce over a { dependencies, files } accumulator so you can do both concats in one assignment, without needing to separate dependencies as a side effect in each of the reduce functions.

also the name of the function is misleading, since it also joins the dependencies.

src/project-generators/react-basic/index.ts
```diff
@@ -1,220 +1,123 @@
...
 
-  // Step 3: Global settings are transformed into the root html file and the manifest file for PWA support
-  if (uidl.globals.manifest) {
-    const manifestJSON = createManifestJSON(uidl.globals.manifest, uidl.name, ASSETS_PREFIX)
-    const manifestFile: GeneratedFile = {
-      name: 'manifest',
-      extension: '.json',
-      content: JSON.stringify(manifestJSON, null, 2),
+  // Step 2: The components generation process is started
+  const componentPromises = Object.keys(components).map(async (componentName) => {
```
vladnicula: should the callback function passed on to the `map` be `async`? Looks like it should not.

src/project-generators/react-next/index.ts
```diff
@@ -1,183 +1,119 @@
...
+  const { states = [] } = root.content
+
+  const stateDefinitions = root.stateDefinitions || {}
+  const routerDefinitions = stateDefinitions.router || null
+
+  // Step 1: The root html file is customized in next via the _document.js page
+  const documentComponentFile = [].concat(createDocumentComponentFile(uidl))
+
+  // Step 2: The first level stateBranches (the pages) transformation in react components is started
+  const pagePromises = states.map(async (stateBranch) => {
```
vladnicula: does not need to be async I think

src/shared/utils/project-utils.ts
```diff
@@ -5,11 +5,17 @@ import {
...
-import { prefixPlaygroundAssetsURL } from '../../shared/utils/uidl-utils'
-import { slugify } from './string-utils'
+} from './html-utils'
 
-export const createHtmlIndexFile = (uidl: ProjectUIDL, assetsPrefix, appRootOverride?: string) => {
+import { prefixPlaygroundAssetsURL, extractPageMetadata } from './uidl-utils'
+import { slugify, sanitizeVariableName } from './string-utils'
+import { FILE_EXTENSIONS } from '../constants'
+
+export const createHtmlIndex = (
```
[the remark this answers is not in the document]
ovidiuionut94: `createHtmlIndexFile` is no longer exported

src/shared/utils/project-utils.ts
```diff
@@ -11,11 +11,15 @@ import { prefixPlaygroundAssetsURL, extractPageMetadata } from './uidl-utils'
 import { slugify, sanitizeVariableName } from './string-utils'
 import { FILE_EXTENSIONS } from '../constants'
 
-export const createHtmlIndex = (
-  uidl: ProjectUIDL,
-  assetsPrefix: string,
+interface HtmlIndexFileParams {
+  uidl: ProjectUIDL
```
alexnm: just a small thing, would it make more sense to have two params to this function? - `projectUIDL` and `options`. it would make the API clearer, because the UIDL is mandatory, all the rest are optional fields that can be sent or not

alexnm:
@allcontributors[bot]  please add @ovidiuionut94 for code

Issue #114, opened 2019-03-28
Explore unified capabilities with respect to our code output


alexnm:
### Current situation:
* we're using HAST (https://github.com/syntax-tree/hast) for representing HTML as we're building the tree.
* HAST builders are in `shared/utils/html-utils`
* we're not using `unified` itself, but an underlying function that is used by **rehype**: `hast-util-to-html`
* the code for converting HAST to HTML is here: `core/builder/generators/html-to-string.ts`

### Objectives:
* revisit doc and check if HAST is a good solution for representing HTML moving further
* avoid doing - hast -> string -> prettier -> ast -> string (try hast -> prettier -> string or hast -> custom format? -> string)
* check performance of the current utility for bigger UIDLs (over 10000 nodes)

### Questions:
* Can we create a single unified pipeline that can handle both HAST and AST?
* Is hastscript - https://github.com/syntax-tree/hastscript - relevant/useful to us?

### References:
* A codesandbox where I played with unified a bit - https://codesandbox.io/s/vj8yv3k1y3
* https://github.com/rehypejs/rehype (including rehype stringify)
* https://unified.js.org/
* https://github.com/prettier/prettier

Issue #96, opened 2019-03-25
Dangerous characters encoding in Vue templates


alexnm:
When generating the `html` for the vue templates, we're using `hast-util-to-html`.

We need to investigate how to bypass the characters encoding for `&`, `'` and other special characters that the vue templates might have inside string attributes.

Example: When using a complex conditional render based on an AND between two operations, we get this
```<span v-if="(activeTab >= 3) &#x26;&#x26; (activeTab < 5)">This is the fourth tab</span>```
instead of this
```<span v-if="(activeTab >= 3) && (activeTab < 5)">This is the fourth tab</span>```

NOTE: The hast-to-html function is documented here: https://github.com/syntax-tree/hast-util-to-html, maybe some properly set options can fix this.

**Comment by JayaKrishnaNamburu** [comment]
Hey @alexnm just need a small help in writing schema's state value for this above ``v-if`` case 😅

**Comment by alexnm** [comment]
try this one for the tab selector states

```
{
  "type": "state",
  "states": [
    {
      "value": 0,
      "content": {
        "type": "container",
        "children": [{
          "type": "text",
          "children": ["This is the first tab"]
        }]
      }
    },{
      "value": 1,
      "content": {
        "type": "text",
        "children": ["This is the second tab"]
      }
    },{
      "value": {
        "matchingCriteria": "any",
        "conditions": [{
          "operand": 2,
          "operation": "==="
        }]
      },
      "content": {
        "type": "text",
        "children": ["This is the third tab"]
      }
    },{
      "value": {
        "matchingCriteria": "all",
        "conditions": [{
          "operand": 3,
          "operation": ">="
        }, {
          "operand": 5,
          "operation": "<"
        }]
      },
      "content": {
        "type": "text",
        "children": ["This is the fourth tab"]
      }
    }
  ]
}
```

The full example is in `__tests__/fixtures/component-sample.json`

JayaKrishnaNamburu:
The config should be like this 
~~~
const unformatedString = toHTML(htmlObject, {
    allowDangerousHTML: true,
    allowDangerousCharacters: true
  })
~~~

alexnm:
I remember trying this but didn't have time to look into it more.

JayaKrishnaNamburu:
ok, I am following up this. I think the repo needs a fix, I was able to achieve this by making some changes to their repo, will update you once I am done 👍

alexnm:
[...]

PR #116, opened 2019-03-29
refactor(html-to-util-string): Switching the HAST to Html builder for…

JayaKrishnaNamburu:
… supporting a few characters

This PR needs a little discussion about the two other issues that are currently open #96  and #114 

Currently, as a test I tried switching from ``hast-util-to-html`` to [@starptech/prettyhtml-hast-to-html](https://github.com/Prettyhtml/prettyhtml/tree/master/packages/prettyhtml-hast-to-html)

The one pro for switching is that we don't need to pass the generated code through prettier, this prettifies the code with the local prettier settings.

@alexnm and @vladnicula your suggestions on this.


alexnm:
@JayaKrishnaNamburu we can switch to the other library as long as we can have the same prettier settings. is this still WIP?

JayaKrishnaNamburu:
> @JayaKrishnaNamburu we can switch to the other library as long as we can have the same prettier settings. is this still WIP?

I am checking why all the tests are breaking but the generators are working fine, will remove the WIP and update here once its done 👍

JayaKrishnaNamburu:
Hey, @alexnm  so here is the update, we need to enforce our prettier configurations because the inbuild configs that they are following are different from the configurations that we need. So, I think we should have our prettier stage as usual

Note: The tests are breaking because of this issue, 

``html-whitespace-sensitive-tag-names`` which is being used as a dev dependency by ``@starptech/prettyhtml-hast-to-html`` that exports as ``json`` but in our jest configurations in ``moduleFileExtensions`` we did not mention ``json`` so the import is not being resolved.

I am new to jest and don't know much on solving this, any little help would be great 😅

alexnm:
@JayaKrishnaNamburu are the example projects running? `create-react-basic`, `create-vue-basic`, etc?

JayaKrishnaNamburu:
Yes they are running, this is because of the jest modulepaths while running test cases, if i add json to the list then tests are passing but again jaon is trying to resolve the schemas too because of this config

alexnm:
can we ignore only specific .json files?

JayaKrishnaNamburu:
Exactly where i am stuck with this, we need module resolver in just one place 😅

alexnm:
we can discuss this maybe during our call later.

JayaKrishnaNamburu:
> we can discuss this maybe during our call later.

Sure, but solved this and updating the branch now. 😄

JayaKrishnaNamburu:
Hi @alexnm please don't merge this, i will let you know in the call

JayaKrishnaNamburu:
So @alexnm  and @vladnicula  here are the outputs which are generated using ``@starptech/prettyhtml-hast-to-html`` So here comes the tradeoff, if we choose to go ahead with this our characters won't be escaped but we will be re-opening issue with boolean states displaying ``true``.

Personally, I would like to test and see with a few other schema's and then observe if we are re-opening or entering into any other edge cases if we plan to choose ``@starptech/prettyhtml-hast-to-htm``

```
<div>
    <div>
      <button @onClick="handleTab1Onclick" class="tab-1">Click</button>
      <button @onClick="handleTab2Onclick" class="tab-2">any</button>
      <button @onClick="handleTab3Onclick" class="tab-3">of</button>
      <button @onClick="handleTab4Onclick" class="tab-4">these</button>
    </div>
    <div class="text">Active tab: {{activeTab}}</div>
    <div v-if="activeTab === 0"><span>This is the first tab</span></div>
    <span v-if="activeTab === 1">This is the second tab</span>
    <span v-if="(activeTab === 2)">This is the third tab</span>
    <span v-if="(activeTab >= 3) && (activeTab < 5)">This is the fourth tab</span>
  </div>
```

```
<div>
    <span>Welcome</span>
    <input />
    <button @onClick="handleDemoButtonOnclick">Demo Button</button>
    <span v-if="toggleComponent">The State of the component is true</span>
    <span v-else="true">This is second</span>
  </div>
```

alexnm:
@JayaKrishnaNamburu is there a different way we can represent `boolean` values as attributes in the HAST representation?

JayaKrishnaNamburu:
> @JayaKrishnaNamburu is there a different way we can represent `boolean` values as attributes in the HAST representation?

Yeah, was looking at that and guess what we can solve it, now we won't be running into this issue too 🚀  😊 , will update the PR in few minutes.

JayaKrishnaNamburu:
Yeah looking at the possibilities will update here 😅  👍

JayaKrishnaNamburu:
So, @alexnm & @vladnicula i just updated this PR, now boolean states (v-if & v-else) are also rendered as we needed. Fixed it by shifting from ``addBooleanAttribute`` to ``addAttributeToNode`` with empty string here --> https://github.com/teleporthq/teleport-code-generators/pull/116/files#diff-34ba58dd94856a83870fa68440f1a7f1R87

src/plugins/teleport-plugin-vue-base-component/utils.ts
```diff
@@ -84,7 +84,7 @@ export const generateVueNodesTree = (
           const alternateNode = getNodeFromContent(alternateContent, accumulators)
           htmlUtils.addAttributeToNode(consequentNode, 'v-if', conditionalStatement)
           htmlUtils.addChildNode(htmlNode, consequentNode)
-          htmlUtils.addBooleanAttributeToNode(alternateNode, 'v-else')
```
alexnm: I propose a different solution to this, please keep `addBooleanAttributeToNode` and modify that function from `project-utils` to add the empty string instead of adding `true`. this way, you solve the problem in all scenarios and we keep the specialized function for adding boolean attributes

JayaKrishnaNamburu: Thought the same thing @alexnm but if we see here ``addBooleanAttributeToNode`` works as the same for a list of predefined HTML attributes reference --> https://github.com/Prettyhtml/prettyhtml/blob/master/packages/prettyhtml-hast-to-html/lib/element.js#L234 by the lib itself, only for the attributes which are not predefined behave in this way.

How about writing a new method and to add empty strings to use in these cases like ``addOverloadBooleanAttributeToNode``? and for all the usual HTML attributes we can use ``addBooleanAttributeToNode``

alexnm: can we import that list of attributes and do a conditional inside our util function? the goal is to provide an API with no friction and no need to know the details about pretty html and which boolean values are allowed

JayaKrishnaNamburu: Yeah @alexnm that will give a smoother experience but if we see here --> https://github.com/Prettyhtml/prettyhtml/blob/master/packages/prettyhtml-hast-to-html/lib/element.js#L225 again they were using another dependency to find out which notes they support, so every time we try to add value to the node they do a cross-check here and then add the value to attribute 😅 

Even if we find the list, then there will check happening two times, one on our side and again on their side 😅  .

JayaKrishnaNamburu: But still, smother experience comes first, I will find something for this today and update here 👍

JayaKrishnaNamburu: Or we can just change our ``addBooleanAttributeToNode`` to add empty string by default so it just wants we needed, here is how they are handeling --> https://github.com/Prettyhtml/prettyhtml/blob/master/packages/prettyhtml-hast-to-html/lib/element.js#L265 so, they would do the two jobs we need.

1. Smoth Dev Experience
2. Just add the attributenames when we want to add a boolean state

Something like this

~~~
export const addBooleanAttributeToNode = (node: HastNode, key: string) => {
  node.properties[key] =''
}
~~~

If we are ok with this, i will make the changes and update the branch 👍

alexnm: go with the empty string please, and leave a comment explaining why `true` doesn't always work

JayaKrishnaNamburu:
@alexnm updated the PR

Issue #35, opened 2019-02-24
Replace $props and $state with proper JSON structure


alexnm:
In the UIDL if we want to address something from props, state, we rely on strings, hence the UIDL manipulation can seem a bit clunky (ex: `$props.message`, `$state.isVisible`)

We could talk about a JSON based structure for defining dynamic content (props, state, local vars - see #10)

A resolution on this should be made before the official release.

vladnicula:
Currently discussing in #87 about a possible UIDL change that addresses this issue as well

vladnicula:
Will be addressed in #87 as working on assets alone is way harder than updating the entire assignment of dynamic values flow

vladnicula:
We should consider changing 

```
"content": {
        "type": "container",
        "name": "container",
        "children": [
          {
            "type": "text",
            "name": "title",
            "children": [...]
          },
```

to:

```
"content": {
        "type": "container",
        "name": "container",
        "children": [
          { 
              type: "component-instance",
              content: {
                "type": "text",
                "name": "title",
                "children": [...]
          }
 }
```

This will allow us to define children of type prop, or state, and not rely on `$props.propName` or `$state.stateName` structure. 

It also brings us closer to having a proper way of differentiating between a instantiation of a component and a declaration.

vladnicula:
Latest update we're trying to implement is this: 

```js
{
   attrs: {
      src: {
            type: "dynamic",
            content: {
                  referenceType:"prop",
                  id: "avatarlURL"
            }
      },
      alt: {
            type: "static",
            content: "the author avatar"
      }
   }
}
```

vladnicula:
@alexnm can we close this? We're done with it right? For beta 0.6 I mean.

PR #120, opened 2019-03-30
refactor(attrs): dynamic and static nodes for all ocurrences of $ str…

alexnm:
…ings

only text nodes are left with the old implementation, but they require more changes in the UIDL
structure. Repeats are now working as expected with both dynamic and static attributes as the
datasource

BREAKING CHANGE: old UIDLs don't work anymore

#35

### DIFF
--- a/src/core/resolver/utils.ts
+++ b/src/core/resolver/utils.ts
@@ -20,6 +20,9 @@ export const mergeMappings = (oldMapping: Mapping, newMapping?: Mapping) => {
   }
 }
 
+// This function is taking a content node instance and returned a resolved version of it,
+// applying the rules set by the mapping structure passed as a parameter
+// The function performs mapping for: attributes, dependencies, repeat, style, events and so on
 export const resolveContentNode = (
   content: ContentNode,
   mapping: Mapping,
@@ -82,13 +85,11 @@ export const resolveContentNode = (
 
       // Data source might be preset on a referenced attribute in the uidl node
       // ex: attrs[options] in case of a dropdown primitive with select/options
-      if (
-        typeof dataSource.content === 'string' &&
-        dataSource.content.startsWith('$attrs.') &&
-        node.attrs
-      ) {
-        const nodeDataSourceAttr = dataSource.content.replace('$attrs.', '')
+      if (dataSource.type === 'dynamic' && dataSource.content.referenceType === 'attr') {
+        const nodeDataSourceAttr = dataSource.content.id
         dataSource = node.attrs[nodeDataSourceAttr]
+
+        // remove original attribute so it doesn't get added as a static/dynamic value on the node
         delete node.attrs[nodeDataSourceAttr]
       }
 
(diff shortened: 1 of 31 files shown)


src/core/resolver/utils.ts
```diff
@@ -81,9 +85,11 @@ export const resolveContentNode = (
 
       // Data source might be preset on a referenced attribute in the uidl node
       // ex: attrs[options] in case of a dropdown primitive with select/options
-      if (typeof dataSource === 'string' && dataSource.startsWith('$attrs.') && node.attrs) {
-        const nodeDataSourceAttr = dataSource.replace('$attrs.', '')
+      if (dataSource.type === 'dynamic' && dataSource.content.referenceType === 'attr') {
```
vladnicula: is this for the special case of mapping different attribute names to other attributes? The one that is used in mappings?

src/plugins/teleport-plugin-react-base-component/utils.ts
```diff
@@ -293,20 +289,29 @@ const makeStateHookAST = (stateIdentifier: StateIdentifier, t = types) => {
...
   const iteratorName = meta.iteratorName || 'item'
   const keyIdentifier = meta.useIndex ? 'index' : iteratorName
 
-  const source =
-    typeof dataSource === 'string'
-      ? t.identifier(dataSource)
-      : t.arrayExpression(dataSource.map((element) => convertValueToLiteral(element)))
+  let source
+
+  if (dataSource.type === 'static') {
```
vladnicula: I suggest we switch by type and throw on default when encountering unknown types. Helps a lot with debugging and TDD.

src/plugins/teleport-plugin-react-base-component/utils.ts
```diff
@@ -293,20 +289,29 @@ const makeStateHookAST = (stateIdentifier: StateIdentifier, t = types) => {
...
+    source = t.arrayExpression(
+      (dataSource.content as any[]).map((element) => convertValueToLiteral(element))
+    )
+  } else {
```
vladnicula: This else implies all types of entities but 'static'. When we add a third one we'll get fuzzy errors here without a default throw error for not recognized type

src/plugins/teleport-plugin-vue-base-component/utils.ts
```diff
@@ -28,7 +28,15 @@ export const generateVueNodesTree = (
 
   if (attrs) {
     Object.keys(attrs).forEach((attrKey) => {
-      addAttributeToNode(htmlNode, name, attrKey, attrs[attrKey], dataObject)
+      const attrValue = attrs[attrKey]
+      // arrays are moved to the data object and referenced as dynamic keys (binding)
+      if (attrValue.type === 'static' && Array.isArray(attrValue.content)) {
```
vladnicula: same as before, maybe we should switch or something and throw if type not recognized.

src/typings/generators.d.ts
```diff
@@ -262,14 +262,14 @@ interface PackageJSON {
...
-    referenceType: 'prop' | 'state' | 'local'
+    referenceType: 'prop' | 'state' | 'local' | 'attr'
     id: string
   }
 }
 
 interface UIDLStaticReference {
   type: 'static'
-  content: string
+  content: string | number | boolean | any[] // array<any> for data sources
```
vladnicula: Why the `any[]` ?

src/core/resolver/utils.ts
alexnm: yes. referenceType = 'attr' will only be handled by the resolver, as it references another attribute in the UIDL node. later on, it might even reference another attribute from another element, but for now it's good enough

src/plugins/teleport-plugin-react-base-component/utils.ts
alexnm: good idea, will change this

PR #124, opened 2019-04-01
Feature/#42 project error handling

anamariaoros:


src/project-generators/react-basic/index.ts
```diff
@@ -78,12 +86,13 @@ const createReactBasicGenerator = (generatorOptions: ProjectGeneratorOptions = {
         componentOptions: {
           assetsPrefix: ASSETS_PREFIX,
           localDependenciesPrefix: LOCAL_DEPENDENCIES_PREFIX,
+          skipValidation: true,
```
ovidiuionut94: What is your opinion of just sending the `skipValidation` flag in the function that creates the page/component outputs? This way you won't have to send it from all the project generator files.

PR #125, opened 2019-04-01
refactor(teleport-plugin-vue-basic): Refractored event handlers for b…

### LINKED ISSUE #95: Event handlers optimization for Vue components
Event handlers are added on the `methods` object in a Vue component. However, when an event handler has a single instruction, like a state assignment, it can be inlined in the template.

Instead of generating:
```
<button @click="handleButtonClick">
```
and adding a function on component methods:
```
methods: {
  handleButtonClick() {
    this.isExpanded = !this.isExpanded
  },
},
```
we could set the state inline
```
<button @click="isExpanded = !isExpanded">
```

JayaKrishnaNamburu:
…oolean states

fixes #95

### DIFF
--- a/src/plugins/teleport-plugin-vue-base-component/utils.ts
+++ b/src/plugins/teleport-plugin-vue-base-component/utils.ts
@@ -34,9 +34,32 @@ export const generateVueNodesTree = (
 
   if (events) {
     Object.keys(events).forEach((eventKey) => {
-      const methodName = `handle${stringToUpperCamelCase(name)}${stringToUpperCamelCase(eventKey)}`
-      methodsObject[methodName] = events[eventKey]
-      htmlUtils.addAttributeToNode(htmlNode, `@${eventKey}`, methodName)
+      if (events[eventKey].length === 1) {
+        const eventDetails = events[eventKey][0]
+        const isPropEvent = eventDetails && eventDetails.type === 'propCall' && eventDetails.calls
+
+        if (isPropEvent) {
+          htmlUtils.addAttributeToNode(
+            htmlNode,
+            `@${eventKey}`,
+            `this.$emit('${eventDetails.calls}')`
+          )
+        } else {
+          htmlUtils.addAttributeToNode(
+            htmlNode,
+            eventDetails.modifies,
+            typeof eventDetails.newState === 'boolean'
+              ? `${eventDetails.modifies} = !${eventDetails.modifies}`
+              : `${eventDetails.modifies} = ${eventDetails.newState}`
+          )
+        }
+      } else {
+        const methodName = `handle${stringToUpperCamelCase(name)}${stringToUpperCamelCase(
+          eventKey
+        )}`
+        methodsObject[methodName] = events[eventKey]
+        htmlUtils.addAttributeToNode(htmlNode, `@${eventKey}`, methodName)
+      }
     })
   }
 
--- a/src/plugins/teleport-plugin-vue-base-component/utils.ts
+++ b/src/plugins/teleport-plugin-vue-base-component/utils.ts
@@ -35,22 +35,18 @@ export const generateVueNodesTree = (
   if (events) {
     Object.keys(events).forEach((eventKey) => {
       if (events[eventKey].length === 1) {
-        const eventDetails = events[eventKey][0]
-        const isPropEvent = eventDetails && eventDetails.type === 'propCall' && eventDetails.calls
+        const statement = events[eventKey][0]
+        const isPropEvent = statement && statement.type === 'propCall' && statement.calls
 
         if (isPropEvent) {
-          htmlUtils.addAttributeToNode(
-            htmlNode,
-            `@${eventKey}`,
-            `this.$emit('${eventDetails.calls}')`
-          )
+          htmlUtils.addAttributeToNode(htmlNode, `@${eventKey}`, `this.$emit('${statement.calls}')`)
         } else {
           htmlUtils.addAttributeToNode(
             htmlNode,
-            eventDetails.modifies,
-            typeof eventDetails.newState === 'boolean'
-              ? `${eventDetails.modifies} = !${eventDetails.modifies}`
-              : `${eventDetails.modifies} = ${eventDetails.newState}`
+            statement.modifies,
+            statement.newState === '$toggle'
+              ? `${statement.modifies} = !${statement.modifies}`
+              : `${statement.modifies} = ${statement.newState}`
           )
         }
       } else {


vladnicula:
@JayaKrishnaNamburu looking good. I have one question though. Did you consider cases like `someValue+=1`. If I understand correctly what you wrote here:  https://github.com/teleporthq/teleport-code-generators/pull/125/files#diff-34ba58dd94856a83870fa68440f1a7f1R38 I think we're only supporting single instruction boolean sets, not increments and decrements.

JayaKrishnaNamburu:
Yes, currently the code I wrote will support only boolean states, will the check the possibility to do for increment and decrement if we need that as well 😄

alexnm:
@JayaKrishnaNamburu similarly with how we handle conditional expression building with JavaScript for event handlers that have more than one instruction, we should develop some helper function for single instructions to generate all possible strings for the one-line event handler.

ex: :onClick="state=value", :onClick="$emit('onClose')"

PS: Check the Modal / ModalWindow example for the `propCall` type of event handler statement

vladnicula:
@JayaKrishnaNamburu if you could also add the increment decrement that would be really great. And if you could move the logic that checks into a utility function it would be perfect

JayaKrishnaNamburu:
Done, will refactor the changes 👍 making all one line actions to follow the same structure for react and vue 👍

JayaKrishnaNamburu:
Hey @alexnm  / @vladnicula do we have any example where we do multiple things in the event, currently is see all of our schema's are having one single statement, I have the following doubts

1. How is it structured, like the event accepts a object or something in the new state ?
2. Are we defining ``increment`` / ``decrement`` in our schemas so we know what kind of operation is going to happen, because if i create a expression lets say ``tab++`` by checking if the new state is number, and what if the user is assigning some value instead of just increment 🤔 

Please help me with this 😅 i am stuck here to proceed on this PR.

vladnicula:
I'm unaware of how events work right now too. Maybe we'll both be enlightened by Alex soon :)

alexnm:
@JayaKrishnaNamburu state management is pretty basic at this point, `$toggle` is the only "complex" function we implemented for convenience. But we will probably open the discussion there as soon as we have the new UIDL structure in place.

`increment`/`decrement` can also be defined as actions that are changing the state. they will always be invoked from event handlers, so inside an UIDLElement, where events are defined we could have:

```
"click": [{
  "type": "stateChange",
   "modifies": "activeTab",
   "newState": 1
},{
  "type": "stateChange,
  "modifies": "activeTab",
  "newState": "2"
}]
```

Doesn't really make sense now, but this would generate:
```
this.activeTab = 1
this.activeTab = 2
```

JayaKrishnaNamburu:
So, currently we can have inline state changes for boolean states and ``propCalls`` and once the further discussion about handling complex states and structure for new UIDL done, we can re-visit this issue and can inline then also right, is my understanding correct regarding this ? 😅

alexnm:
yes, we will revisit this. but currently (in v0.5) standard assignment also works. check the code generated for the TabSelector in both Vue and React. you have `activeTab = 1` and `setActiveTab(1)` for Vue and React respectively

JayaKrishnaNamburu:
Cool, i will inline the ``propCalls`` for react and vue and we are good to go with this PR before the discussion on advanced used cases then :) Thank you 😄 👍

JayaKrishnaNamburu:
Made prop event too as inline in vue-generators as below, @alexnm please review the PR, will handle the increment and decrement once we have a discussion on handling this, react-generators are already following this style of representation. Please let me know if i miss out anything 😅 
```
<button @click="handleCloseClick">Close me</button>
```

```
const ModalWindow = (props) => {
  return (
    <div>
      <span>{props.message}</span>
      <button onClick={props.onClose}>Close me</button>
    </div>
  )
}
```

src/plugins/teleport-plugin-vue-base-component/utils.ts
```diff
@@ -34,9 +34,31 @@ export const generateVueNodesTree = (
 
   if (events) {
     Object.keys(events).forEach((eventKey) => {
-      const methodName = `handle${stringToUpperCamelCase(name)}${stringToUpperCamelCase(eventKey)}`
-      methodsObject[methodName] = events[eventKey]
-      htmlUtils.addAttributeToNode(htmlNode, `@${eventKey}`, methodName)
+      const eventDetails = events[eventKey][0]
+      const isPropEvent = eventDetails && eventDetails.type === 'propCall' && eventDetails.calls
+      const isBooleanChangeEvent =
```
alexnm: you can also do one line handlers for non-boolean change events. please check the example with the tab selector, where 4 different handlers are created for the 4 buttons that are performing the tab selection

alexnm: basically, this function should be:

if - one statement - render propCall or stateChange as string
if - more statements - render eventHandlerByName, add method name to `methodsObject `

JayaKrishnaNamburu: Sure, will make the changes 👍

JayaKrishnaNamburu:
Hey @vladnicula check this once !!

src/plugins/teleport-plugin-vue-base-component/utils.ts
```diff
@@ -34,9 +34,32 @@ export const generateVueNodesTree = (
...
+          htmlUtils.addAttributeToNode(
+            htmlNode,
+            `@${eventKey}`,
+            `this.$emit('${eventDetails.calls}')`
+          )
+        } else {
+          htmlUtils.addAttributeToNode(
+            htmlNode,
+            eventDetails.modifies,
+            typeof eventDetails.newState === 'boolean'
```
alexnm: condition should be `typeof eventDetails.newState === '$toggle'` the same as it is implemented in the AST generator for the JS event handler

src/plugins/teleport-plugin-vue-base-component/utils.ts
```diff
@@ -34,9 +34,32 @@ export const generateVueNodesTree = (
 
   if (events) {
     Object.keys(events).forEach((eventKey) => {
-      const methodName = `handle${stringToUpperCamelCase(name)}${stringToUpperCamelCase(eventKey)}`
-      methodsObject[methodName] = events[eventKey]
-      htmlUtils.addAttributeToNode(htmlNode, `@${eventKey}`, methodName)
+      if (events[eventKey].length === 1) {
+        const eventDetails = events[eventKey][0]
```
alexnm: also forgot to mention, `eventDetails` can be renamed here as `statement`

src/plugins/teleport-plugin-vue-base-component/utils.ts
JayaKrishnaNamburu: Sorry, i mis took all boolean states will behave as toggle's, changed now 👍

PR #127, opened 2019-04-01
Refactor/#117 style props uidl

### LINKED ISSUE #117: Style assignments UIDL refactor
Related to #105

Same as attributes, style can have assigned strings, numbers, and dynamic values. The same interface can be used read and handle props, state, local values in style props.

A special case that needs to be tested is the situation when a style object has nested objects, in which case we need to introduce maybe another type of assignment, or to extend the static/dynamic assignments to entire objects.

vladnicula:
fix #117


__tests__/integration/component-assignemnts/component-with-invalid-style.json
```diff
@@ -0,0 +1,18 @@
+{
```
alexnm: you have a typo in the folder name, `component-assignemnts` :)

src/shared/utils/uidl-utils.ts
```diff
@@ -82,82 +82,124 @@ export const traverseNodes = (node: ContentNode, fn: (node: ContentNode) => void
...
-        dynamicStyles[key] = nestedResult.dynamicStyles
-      }
-      if (Object.keys(nestedResult.staticStyles).length > 0) {
-        staticStyles[key] = nestedResult.staticStyles
-      }
-    } else if (typeof value === 'string' && value.startsWith('$props.')) {
-      dynamicStyles[key] = value
-    } else {
-      staticStyles[key] = value
+  Object.keys(style).reduce((acc: SplitResponse, styleKey) => {
```
alexnm: since we're returning two separate structures, does it help that we use reduce here? a forEach seems more straight forward to parse the style object

Issue #118, opened 2019-03-29
Add method of transforming string based props and style to JSON structure


vladnicula:
We need a utility function that could potentially be exposed as a "adaptor" plugin in the pipeline or generators which would transform a UIDL that has attributes and style like:

```js
{
  styleORAttrs: {
    href: "url"
  }
}
```
into

```
{
  styleORAttrs: {
    href: { type: "static", "content": "url"
  }
}

**Comment by vladnicula** [comment]
The second type of transformation is:
```
{
  styleORAttrs: {
    href: "$props.url"
  }
}
```

into

```
{
  styleORAttrs: {
    { type: "props", "content": { id: "url" } }
  }
}
```

vladnicula:
We now have in release/0.6 a couple of utility functions that could be used to transform old UIDL into new one. 

The missing functionality as of now is the following:
- Children transformations as @alexnm is still working on the structure of children (and specifically, state based children) 
- A unified way of transforming, probably before the uidl is fed into the project/component generators. This will require a composition of the new utility functions added in the last MR

vladnicula:
We are waiting on #143 to wrap up before finalising this.

PR #143, opened 2019-04-03
feat(uidl-schema): uidl-schema-update

anamariaoros:


src/uidl-definitions/schemas/new-component.json
```diff
@@ -0,0 +1,279 @@
...
+                    }
+                  }
+                },
+                "defaultValue": {},
+                "actions": {"type": "array"}
+              }
+            }
+          }
+        },
+        "style": {},
```
alexnm: just a quick note, style could also be now expanded into `<any-key>`: staticValue / dynamicReference + nested styles, you can talk with @vladnicula about the structure tomorrow

src/uidl-definitions/schemas/new-component.json
```diff
@@ -0,0 +1,279 @@
...
+        "node": {
+          "oneOf": [{
```
alexnm: this oneOf is good, but we could also add a separate `string` as a type, if we want to support the old style text nodes, ex: "$props.title". let's talk tomorrow

src/uidl-definitions/schemas/new-component.json
```diff
@@ -0,0 +1,279 @@
...
+          "type": "object",
+          "additionalProperties": false,
+          "required": [ "type", "content" ],
+          "properties": {
+            "type": {
+              "type": "string",
+              "enum":  ["repeat"]
+            },
+            "content": {
+              "$ref": "#/definitions/repeatContent"
```
alexnm: no need to split this if it's not reused somewhere else

vladnicula: I'd argue we could split it for ease of editing. I'd rather have the content of the repeat in a separate file for example and only focus on it if we make changes in the future.

vladnicula:
Will need this merged before integrating #118

Issue #41, opened 2019-02-24
Extend the UIDL validator with business rules


alexnm:
UIDL validation is currently performed strictly on the JSON schema structure.

We could add a thin layer of validation for:
- prop that is being used without being defined in `propDefinitions`
- prop that is defined but not used in the content area
- state that is being used without being defined in `stateDefinitions`
- state that is defined but not used in the content area
- prop and state key are not the same? (this could work in React, but not in Vue)
- using local variables inside repeat:
   * using index without declaring `useIndex` in meta
   * using custom local variable name without specifying it in meta as `iteratorName`

Additionally for project generators:
- component is referenced but does not exist in the project
- component name is different than component key (in project UIDLs)
- external dependency version is consistent across the project
- `route` state key is defined inside `root`
- the first level children in the `root` component are conditionals based on `route`

alexnm:
this is dependent on the implementation of the validation layer in #2

JayaKrishnaNamburu:
Hey @alexnm is this available, i don't have much experience with resolvers, but would like to give a try and complete this 😃

alexnm:
@JayaKrishnaNamburu I would wait for #140 to be merged into master. or you can start from that branch if you'd like. but this issue will require some changes after all the things we did for v0.6.

JayaKrishnaNamburu:
Oh, ok i will wait for the master to be stable with the new changes and start with this, after a discussion for new changes instead of twice 😅

JayaKrishnaNamburu:
Hey @alexnm since the 0.6 is merged can i start on this ?

alexnm:
hey, I have a task to update this with the props specs, so not yet

JayaKrishnaNamburu:
Ok, here are the few observations that we need to take care when doing this PR

1. #145 Misspelling of state identifier crashes the component generator
2. There is an edge case with all the style flavors. Basically, we are not doing a cross-check when the user is trying to refer a prop or state, we are generating the code. But when we try to run that only then we will know that the user did not declare them.

PR #147, opened 2019-04-04
WIP: fix(teleport-react & teleport-vue basic plugins): Continuing the comp…

### LINKED ISSUE #145: Missing/ Mispelling a state key crashes the component generation
# Version 0.6

Hey, I was just testing the new UIDL and component generations, here is something i noticed
~~~
              "content": {
                "reference": {
                  "type": "dynamic",
                  "content": {
                    "referenceType": "state",
                    "id": "isVisible"
                  }
                },
~~~

Let us take the above example if i am miss-spelling or not mentioning ``isVisible`` it would be so useful if we skip the generation of that particular statement instead of blocking the whole component generation. Currently if we miss this attribute or wrongly spell that, the component is not being generated as we are trying to find the element in state.

JayaKrishnaNamburu:
…onent generation process by ski

fixes #145

### DIFF
--- a/src/plugins/teleport-plugin-react-base-component/utils.ts
+++ b/src/plugins/teleport-plugin-react-base-component/utils.ts
@@ -95,15 +95,21 @@ export const generateConditionalNode = (
   accumulators: ReactComponentAccumulators
 ) => {
   const { reference, value } = node.content
-  const conditionIdentifier = createConditionIdentifier(reference, accumulators)
+  const stateKey = reference.content.id
+  const stateIdentifier = accumulators.stateIdentifiers[stateKey]
+  
+  if (stateIdentifier) {
+    const subTree = generateNodeSyntax(node.content.node, accumulators)
 
-  const subTree = generateNodeSyntax(node.content.node, accumulators)
+    const condition: UIDLConditionalExpression = value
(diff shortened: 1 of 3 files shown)


src/plugins/teleport-plugin-react-base-component/utils.ts
```diff
@@ -96,14 +96,21 @@ export const generateConditionalNode = (
 ) => {
   const { reference, value } = node.content
   const conditionIdentifier = createConditionIdentifier(reference, accumulators)
+  const stateKey = reference.content.id
+  const stateIdentifier = accumulators.stateIdentifiers[stateKey]
```
JayaKrishnaNamburu: Hey @alexnm i am thinking if i should stop implementing this additional check in generators and close the PR, because when i am developing i felt if we are doing a check for state, we should do for prop definitions as well and then the check are increasing, As you mentioned here --> https://github.com/teleporthq/teleport-code-generators/issues/145#issuecomment-479809748 now even i fell i might be adding additional checks into the generators if the #41 is implemented. What you say ?

alexnm: @JayaKrishnaNamburu ok, let's close this as we have plenty of changes in release/0.6 already

PR #150, opened 2019-04-04
Fix for #148

### LINKED ISSUE #148: React dynamic children don't add props.NAME_OF_PROP
this UIDL child form a `node.children[]`
```json
 {
     "type": "dynamic",
     "content": {
          "referenceType": "prop",
         "id": "title"
    }
}
```
produces the following sytanx.
```jsx
<span>
        Hello World!
        {title}
</span>
```

The problem is that title needs to be prefixed by `props.` like

```jsx
<span>
        Hello World!
        {props.title}
</span>
```
We should add a test for this as well.

vladnicula:
Fix #148

### DIFF
--- a/src/plugins/teleport-plugin-react-base-component/utils.ts
+++ b/src/plugins/teleport-plugin-react-base-component/utils.ts
@@ -117,7 +117,7 @@ export const generateNodeSyntax: NodeSyntaxGenerator<
       return node.content.toString()
 
     case 'dynamic':
-      return makeDynamicValueExpression(node.content.id)
+      return makeDynamicValueExpression(node)
 
     case 'element':
       return generateElementNode(node, accumulators)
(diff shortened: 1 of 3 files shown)


src/plugins/teleport-plugin-react-base-component/utils.ts
```diff
@@ -334,8 +334,11 @@ const makeRepeatStructureWithMap = (
   )
 }
 
-const makeDynamicValueExpression = (identifier: string, t = types) => {
-  return t.jsxExpressionContainer(t.identifier(identifier))
+const makeDynamicValueExpression = (identifier: UIDLDynamicReference, t = types) => {
+  const prefix = getReactVarNameForDynamicReference(identifier)
```
alexnm: does this work if prefix is empty?

PR #155, opened 2019-04-05
Feature/#118 uidl transform in pipeline

vladnicula:
This finalises #118


__tests__/unit/core/parser/component-with-proper-values.json
```diff
@@ -0,0 +1,98 @@
...
+
+      "style": {
+        "flexDirection": {
+          "type": "dynamic",
+          "content": {
+            "referenceType":"prop",
+            "id": "test"
+          }
+        },
+        "alignSelft": {"type":"static", "content": "center"}
```
alexnm: alignSelf :)

src/core/parser/component.ts
```diff
@@ -0,0 +1,50 @@
...
+
+      if (elementContent.attrs) {
+        elementContent.attrs = transformAttributesAssignmentsToJson(elementContent.attrs as Record<
+          string,
+          unknown
+        >)
+      }
+
+      if (Array.isArray(elementContent.children)) {
+        elementContent.children = elementContent.children.reduce(
```
alexnm: you can use a `map` instead of `reduce`, as it is doing a 1:1 transform of the array

src/core/parser/component.ts
```diff
@@ -0,0 +1,50 @@
...
+          (newChildren: UIDLNode[], child) => {
+            if (typeof child === 'string') {
+              newChildren.push(transformStringAssignmentToJson(child))
+            } else {
+              newChildren.push(parseComponentJSON({ node: child }).node)
```
alexnm: why not parseComponentJSON(child) ?

src/core/parser/component.ts
```diff
@@ -0,0 +1,50 @@
+import {
+  transformStylesAssignmentsToJson,
+  transformAttributesAssignmentsToJson,
+  transformStringAssignmentToJson,
+} from '../../shared/utils/uidl-utils'
+
+import { ComponentUIDL, UIDLNode } from '../../typings/uidl-definitions'
+
+export const parseComponentJSON = (input: Record<string, unknown>): ComponentUIDL => {
+  const node = input.node as Record<string, unknown>
```
alexnm: should we declare this Record<string, unknown> as RawInput or something like that? if we start using it in a lot of places it will be some repetition

src/core/parser/project.ts
```diff
@@ -0,0 +1,22 @@
+import { parseComponentJSON } from './component'
+
+import { ProjectUIDL, ComponentUIDL } from '../../typings/uidl-definitions'
+
+export const parseProjectJSON = (input: Record<string, unknown>): ProjectUIDL => {
+  const root = input.root as ComponentUIDL
+  const components = (input.components || {}) as Record<string, ComponentUIDL>
+
+  const result = {
```
alexnm: the resolver uses a `cloneObject` function at some point to avoid reference leaking and to ensure the input UIDL is not modified by accident. should we move that here now that this is the first step where we potentially change the input?

src/core/parser/component.ts
vladnicula: oh good point, this evolved into the current api and I didn;t get to change it.

src/core/parser/project.ts
vladnicula: will do

src/core/parser/component.ts
vladnicula: Seems a bit generic and not worth the import imo :-?

src/core/parser/component.ts
vladnicula: will do

__tests__/unit/core/parser/component-with-proper-values.json
vladnicula: ah maaan, I had like 20 of these. Some slipped the mass rename.

src/core/parser/component.ts
vladnicula: actually because `parseComponentJSON({ node: child })` returns a component without anything but a node key. parseComponentJSON(child) is invalid because child is a `UIDLnode`, not a `UIDLComponent `

src/core/parser/project.ts
vladnicula: It's unclear if we should clone in components as well as in the project. If I add a clone in both place, we will clone twice, once in project and once in component 👎

__tests__/unit/core/parser/component-with-proper-values.json
alexnm: :))) another minus for JSON representation

src/core/parser/component.ts
alexnm: so it will remain {node: child} ?

I was asking about `{node: child}.node` which is equivalent with `child`

src/core/parser/project.ts
alexnm: maybe we can put it behind the skipValidation flag? or use a skipParsing flag?

parsing is in a way part of the pre-processing of the UIDL so we can use a single flag for both. no idea which is the best option for now

vladnicula:
Any thoughts on `noClone` params? Should we call the component generators with it set to true in project generators? 

Shall we remove the clone from the resolver entirely?

src/core/parser/component.ts
```diff
@@ -0,0 +1,102 @@
...
+}
+
+export const parseComponentJSON = (
+  input: Record<string, unknown>,
+  params: ParseComponentJSONParams = {}
+): ComponentUIDL => {
+  const safeInput = params.noClone ? input : cloneObject(input)
+
+  const node = safeInput.node as Record<string, unknown>
+  const result: ComponentUIDL = {
```
alexnm: I think result can be assigned directly on the first line and then be cast to ComponentUIDL at the end when it is returned

src/core/parser/project.ts
```diff
@@ -0,0 +1,30 @@
...
+  params: ParseProjectJSONParams = {}
+): ProjectUIDL => {
+  const safeInput = params.noClone ? input : cloneObject(input)
+  const root = safeInput.root as ComponentUIDL
+
+  const result = {
+    ...(safeInput as ProjectUIDL),
+  }
+
+  result.root = parseComponentJSON(root)
```
alexnm: will you pass `noClone` as a param at this point?

src/core/parser/component.ts
vladnicula: I need the type to be defined before adding anything on `results.node` for it to run typescript and warn if I am trying to assign an invalid value to node.

src/core/parser/project.ts
vladnicula: good point

PR #158, opened 2019-04-06
feat(slots): support for React children and Vue <slot>

### LINKED ISSUE #156: Support for slots
Originally we were supporting the concept of `props.children` from React, but we believe that slots are a bit closer to the web components standard.

With the new UIDL structure, we will be able to write something like this:

```{ "type": "dynamic", "content": { "referenceType": "slot", "id": "default" }```

This will be translated into

```{ props.children }``` in React
```<slot></slot>``` in Vue

### PR DESCRIPTION
**Pull request by alexnm**
closes #156

### CODE REVIEW

File: __tests__/integration/component-slots/index.ts [hunk 1]
```diff
@@ -0,0 +1,73 @@
...
+    const uidl = component(
+      'Slot Component',
+      elementNode('container', {}, [slotNode(undefined, 'slot-1')])
+    )
+
+    it('renders props.children in React', async () => {
+      const result = await reactGenerator.generateComponent(uidl)
+      const code = result.files[0].content
+
+      expect(code).toContain('{props.children}')
```
vladnicula: If `slot-1` is a named slot, the the expected props to be sent over to react would  be `props['slot-1']`. Otherwise react would not have feature parity with vue. Vue will be able to have header, footer and main content as 3 separate slots configurable, while react will only support one.

Components like layouts (header, content, footer) need this in my opinion.

src/plugins/teleport-plugin-react-base-component/utils.ts
```diff
@@ -124,6 +125,38 @@ export const generateConditionalNode = (
   return createConditionalJSXExpression(subTree, condition, conditionIdentifier)
 }
 
+export const generateSlotNode = (node: UIDLSlotNode, accumulators: ReactComponentAccumulators) => {
+  const childrenProp: UIDLDynamicReference = {
+    type: 'dynamic',
+    content: {
+      referenceType: 'prop',
+      id: 'children',
```
vladnicula: this is probably the place where a slot with default name becomes props.children while a named slots becomes props.nameOfSlot.

src/plugins/teleport-plugin-react-base-component/utils.ts
```diff
@@ -124,6 +125,38 @@ export const generateConditionalNode = (
...
+    },
+  }
+
+  const childrenExpression = makeDynamicValueExpression(childrenProp)
+
+  if (node.content.fallback) {
```
vladnicula: was this needed for vue only? In react default props would already provide this kind of functionality, right?

src/plugins/teleport-plugin-react-base-component/utils.ts
```diff
@@ -124,6 +125,38 @@ export const generateConditionalNode = (
...
+    if (typeof fallbackContent === 'string') {
+      expression = types.stringLiteral(fallbackContent)
+    } else if ((fallbackContent as types.JSXExpressionContainer).expression) {
+      expression = (fallbackContent as types.JSXExpressionContainer).expression as types.Expression
+    } else {
+      expression = fallbackContent as types.JSXElement
+    }
+
+    // props.children with fallback
+    return types.jsxExpressionContainer(
```
vladnicula: are we drooping the optional last param `t=types` form the functions now? If yes why, and should we remove all the optional params t=types now?

src/shared/utils/uidl-utils.ts
```diff
@@ -101,6 +101,10 @@ export const traverseNodes = (
   if (node.type === 'conditional') {
     traverseNodes(node.content.node, fn, node)
   }
+
+  if (node.type === 'slot' && node.content.fallback) {
```
vladnicula: we should make a switch here as we never run two if statements in the same function call for traverseNode. We also don't capture errors for unsupported types, right?

src/plugins/teleport-plugin-react-base-component/utils.ts
alexnm: no, I think we can keep it, I missed that here, I'll add it. but maybe we should isolate at some point babel/types usage to some specific utils/builders modules so we can test the other layers without worrying about this.

src/shared/utils/uidl-utils.ts
alexnm: yes, will change both traverseNodes and traverseElements

__tests__/integration/component-slots/index.ts
alexnm: I was wondering if we need some sort of abstraction for the slots when passing them.

In Vue, there's a standard where you add `v-slot:header` `v-slot:footer` as attributes, so we wouldn't be able to use the same UIDL for vue/react if we don't have some sort of abstraction on that end

PR #160, opened 2019-04-06
Fix/npm pack size

### LINKED ISSUE #159: Reducing the package size of our npm package
## Improvements 🚀 

Hello, so as a front end developer i always like to make my bundls less in size so i can cheat on compile and parse times of the bundle. So today when i am trying to download our package from **npm** i observed we are packing few things that are actually not needed for the end users to use in their projects, here are few of my observations.

1. We are exposing our entry point as **dist/index.js** but still we are packing our src too in the pack.
2. We are packing our ``uidl-samples`` which were actually in-accessible when we install teleport as a dependency in any other project.
3. This is also taking a sum of nearly ``103kb`` in the package, we have a ``png`` image of the ``teleporthq`` logo. (Here is an awesome [article](https://medium.com/s/silicon-satire/i-peeked-into-my-node-modules-directory-and-you-wont-believe-what-happened-next-b89f63d21558) that talk about this)

I made few changes and added ``.npmignore`` and generated a package with the new changes and the ``gzip`` size is **103kb** now.

Here are the [files](https://drive.google.com/drive/folders/1xD2xA9V7B_ltqJGJ6O_t2H7y8FsYbugb?usp=sharing), one is from the npm and the other is locally generated, and here is how to test it. After downloading the file, move it to the folder where you want to install and then yu can directly run ``npm install <file-name>``

This is how your ``package.json`` will look
~~~
"@teleporthq/teleport-code-generators": "file:teleporthq-teleport-code-generators-0.5.1.tgz"
~~~

Please try this and let me know your views and suggestions, we can bring the size from **298kb** to **103kb**  

# 🔥

JayaKrishnaNamburu:
fixes #159

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,37 @@
+__tests__
+.idea
+.DS_STORE
+*~
+*.pyc
+build/
+coverage/
+*.log*
+*.sublime-project
+*.sublime-workspace
+.idea
+*.iml
+.vscode
+*.swp
+*.swo
+
+# Logs
+logs
+*.log
+npm-debug.log*
+
+# Coverage directory used by tools like istanbul
+coverage
+
+# Compiled binary addons (https://nodejs.org/api/addons.html)
+build/Release
+
+# Dependency directories
+node_modules/
+
+# Output of 'npm pack'
+*.tgz
+
+examples/uidl-samples/
+
+# Any media files
+*.png


alexnm:
@JayaKrishnaNamburu I had a discussion with @Utwo and @vladnicula and we'd like to keep the `src` folder on npm so that everyone can look into the code inside `node_modules` when they use the generators. Also, everything that is in `.gitignore` will be ignored by `npm publish` as well so there's no need to add it here.

alexnm:
I would close this if no one has anything else to add

JayaKrishnaNamburu:
Currently, the only concern is the image, if we could just remove the image from npm pack, that will alone remove nearly 100-110kb of size, rest of all iterations like src, and files we can discuss when we are focusing on shifting to Lerna 😄 or we can just take care of the whole thing there itself and we can close this 👍 since the npm publish totally changes once we adopt lerna 👍

alexnm:
@JayaKrishnaNamburu makes sense, please keep the file then with just the image ignore and let's merge this!

JayaKrishnaNamburu:
> @JayaKrishnaNamburu makes sense, please keep the file then with just the image ignore and let's merge this!

Sure, will update the PR right away 👍

JayaKrishnaNamburu:
@alexnm just removed ``src`` from the file, it turns out if ``.npmignore`` is preset ``.gitignore`` is not showing any effect on the packed size, the other alternative is to specify all the files that are needed in ``packcage.json`` explicitly.

alexnm:
not sure I understand. so if .npmignore is present, .gitignore is "ignored" by `npm publish`?

JayaKrishnaNamburu:
Yes, in my local i am noticing that behavior

with the current ``.npmignore`` file i am having an output something like this
~~~
npm notice === Tarball Details === 
npm notice name:          @teleporthq/teleport-code-generators         
npm notice version:       0.6.0                                        
npm notice filename:      teleporthq-teleport-code-generators-0.6.0.tgz
npm notice package size:  152.1 kB                                     
npm notice unpacked size: 738.1 kB                                     
npm notice total files:   324                                          
~~~

By keeping only png in the ``.npmignore``
~~~
npm notice === Tarball Details === 
npm notice name:          @teleporthq/teleport-code-generators         
npm notice version:       0.6.0                                        
npm notice filename:      teleporthq-teleport-code-generators-0.6.0.tgz
npm notice package size:  497.6 kB                                     
npm notice unpacked size: 2.5 MB                                       
npm notice total files:   457  
~~~

JayaKrishnaNamburu:
> not sure I understand. so if .npmignore is present, .gitignore is "ignored" by `npm publish`?

yes, here are the outputs ☝️

JayaKrishnaNamburu:
@alexnm  here it is --> https://stackoverflow.com/questions/24942161/does-npm-ignore-files-listed-in-gitignore looks like if ``.npmignore`` is present, the ``npm pack`` gives priority only to that file, only in case if ``.npmignore`` is missing then it picks ``gitignore``

alexnm:
@JayaKrishnaNamburu I would not make any changes for now until we have the lerna or other project structure setup. Perhaps we will have a separate npmignore in each package after the split. That would work well and will not bring any duplication. I will close this for now. Please reopen it if you feel like there's something needed to be merged at this stage.

PR #167, opened 2019-04-09
Tests/code coverage increase

alexnm:


src/uidl-definitions/schemas/component.json
```diff
@@ -220,6 +220,7 @@
     "elementNodeContent": {
       "type": "object",
       "required": ["elementType"],
+      "additionalProperties": true,
```
vladnicula: what additional keys do we need on the element node content?

PR #191, opened 2019-04-17
refactor(teleport-react-base-component): Throwing error when the root…

### LINKED ISSUE #182: Handle non-element root for React component generator
With the new node based implementation for the UIDL structure, it's possible to define non-elements as the first node inside a component.

For example a component could start with a conditional node, a slot, or a repeat. It might not be a good practice, but we have to make sure we support all UIDL combinations.

The result of the `generateNodeSyntax` inside the `react-base` plugin is being cast to `JSXElement`:

```
const jsxTagStructure = generateNodeSyntax(uidl.node, {
      propDefinitions: uidl.propDefinitions || {},
      stateIdentifiers,
      nodesLookup,
      dependencies,
    }) as types.JSXElement
```

We need to handle other types of results her correctly

JayaKrishnaNamburu:
… node is conditional

fixes #182

### DIFF
--- a/src/plugins/teleport-plugin-react-base-component/index.ts
+++ b/src/plugins/teleport-plugin-react-base-component/index.ts
@@ -21,6 +21,10 @@ export const createPlugin: ComponentPluginFactory<JSXConfig> = (config) => {
   const reactComponentPlugin: ComponentPlugin = async (structure) => {
     const { uidl, dependencies } = structure
 
+    if (uidl.node.type === 'conditional') {
+      throw new Error('Please check the UIDL structure, Root element cannot be a conditional')
+    }
+
     dependencies.React = {
       type: 'library',
       path: 'react',


alexnm:
Let's talk about this on Monday, we were quite busy with the team retreat these past 3 days and I have a couple of things I want to check related to this particular issue

src/plugins/teleport-plugin-react-base-component/index.ts
```diff
@@ -21,6 +21,10 @@ export const createPlugin: ComponentPluginFactory<JSXConfig> = (config) => {
   const reactComponentPlugin: ComponentPlugin = async (structure) => {
     const { uidl, dependencies } = structure
 
+    if (uidl.node.type === 'conditional') {
```
vladnicula: Similar things happen when a repeat is the root node. :-? I'm not sure if just a 'conditional' should throw an error. I'm wondering if we could actually have conditionals and repeats supported as roots instead of just elements :-?

JayaKrishnaNamburu: Yeah, i realised after giving the PR, that it is possible to have ``repeat`` as the root element, but my personal opinion is it would be good if we can support both repeat and conditional as root, by wrapping the component with ``Fragments`` in this specific case, in that way we won't be running into this. Alex said he will be looking into this, let's see 👍

alexnm:
@JayaKrishnaNamburu I would close this until we do the lerna setup, to not have conflicts here and there, I'd take the next few days to focus on that one first. I will update the issue (#182) with what I find later this week

PR #194, opened 2019-04-22
Feature/lerna

JayaKrishnaNamburu:
Hi, @alexnm and @vladnicula you can use this branch for testing lerna setup and other stuff.

Things that are currently done
1. Splitting into multiple packages.
2. We can send css plugins whenever needed.

Things need to be done
1. Way to refactor all test cases.
2. Setting up changelog management for all Individual packages.
3. Cleanup of readme files, if we adopt this approach

Will update all the links and steps to set up this branch and linking packages in a short while.


JayaKrishnaNamburu:
Steps to work with the setup

1. ``lerna create <package_name>`` will create a new package with the name that is specified.
2.  Edit the entry points and add a build script, which we use when during the building phase.
3. The package is created under a folder called ``packages`` which we configured in ``lerna.json`` file
4. For example, we created a package called ``A`` and a package called ``B`` and we want to add ``B`` as a dependency for ``A``, simply add the package ``B`` in ``dependencies`` of package ``A`` and then run ``yarn install`` in the packages folder, this step will create sys links between the packages, so the changes done in one package will reflect in another one.
5. Since we are using typescript, but at the end the machines can understand javascript, when we make a change to a specific package, we need to run ``yarn build`` in that package so it updates the ``dist`` folder in the package, which in turn is exposed outside.
6. ``lerna run build`` will invoke the ``build`` script which we specified in all the packages concurrently.
7. ``lerna publish`` is used to tag the release and push the packages to the ``npm`` for this we need to login under the ``package-namespace`` like in our case, we should login under ``@teleporthq`` organization, only then all the packages are pushed under the organization.
8. After development we might want to do a round of internal testing to check if the packages that are published are working correctly, we need to way to publish them, for this we cannot use the ``npm registry`` directly, so we can use [verdaccio](https://github.com/verdaccio/verdaccio) which works as a proxy registry when we can publish and download the packages to check if they are working properly.

So, when we run ``lerna publish`` we need to select the kind of release, which means a patch, pre-release, minor, major etc.. when we select an option depending on that all the dependent packages are updated, for example in the above scenario we were using package ``B`` as dependency for ``A`` so when we publish the version number of ``B`` in its ``package.json`` and in ``A``'s ``package.json`` are updated.

lerna.json
```diff
@@ -0,0 +1,14 @@
+{
+  "packages": [
+    "packages/*"
+  ],
+  "version": "independent",
```
JayaKrishnaNamburu: @alexnm we need to a dummy git publish in ``independent`` mode before adopting, today I was experimenting with ``independent`` mode and lerna pushed all packages as individual releases in my fork.

alexnm: what do you mean by git publish? npm?

JayaKrishnaNamburu: sorry, i mean Releases tab in github :sweat_smile:

packages/teleport-component-generator-react/package.json
```diff
@@ -0,0 +1,32 @@
+{
+  "name": "@teleporthq/teleport-component-generator-react",
+  "version": "0.7.0",
```
JayaKrishnaNamburu: @alexnm i ideal way will be we keep all our generators as 0.6 and then we use ``lerna publish`` so it will bump to 0.7 and push to npm and tag the releases. In this way if we do a publish it will bump to 0.8 or a path or prerelease or something and then it pushes them. We can do a test run with this with verdaccio and confirm

alexnm: feel free to update the version directly on this branch. I will check it out later as I'm not at my laptop now

packages/teleport-component-generator-react/src/react-component.ts
```diff
@@ -1,9 +1,10 @@
 import { AssemblyLine, Builder, Resolver, Validator } from '@teleporthq/teleport-generator-core'
 
 import reactComponentPlugin from '@teleporthq/teleport-plugin-react-base-component'
-import reactInlineStylesPlugin from '../../teleport-plugin-react-inline-styles/lib'
-import reactPropTypesPlugin from '../../teleport-plugin-react-proptypes/lib'
-import importStatementsPlugin from '../../teleport-plugin-import-statements/lib'
+import reactInlineStylesPlugin from '@teleporthq/teleport-plugin-react-inline-styles'
+import reactStyledJSXPlugin from '@teleporthq/teleport-plugin-react-styled-jsx'
+import reactPropTypesPlugin from '@teleporthq/teleport-plugin-react-proptypes'
```
JayaKrishnaNamburu: @alexnm if we bundle this way, we may end up packing all the supported plugins again in the component-generator right? which in returns effects the bundle size, in that way we might lose the power of splitting the plugins into submodules. 🤔

alexnm: it's a temporary solution, the idea is to have feature parity with the old style monorepo, then we start working on dynamic plugins

package.json
```diff
@@ -1,30 +1,27 @@
...
     "commitizen": "git-cz",
     "prepare": "npm run build",
+    "clean": "lerna run clean",
     "add-contributor": "all-contributors add",
-    "create-react-next": "TS_NODE_FILES=true ts-node --project tsconfig.json ./examples/project-exporters/react-next/index.ts",
-    "create-react-basic": "TS_NODE_FILES=true ts-node --project tsconfig.json ./examples/project-exporters/react-basic/index.ts",
-    "create-vue-basic": "TS_NODE_FILES=true ts-node --project tsconfig.json ./examples/project-exporters/vue-basic/index.ts",
-    "create-vue-nuxt": "TS_NODE_FILES=true ts-node --project tsconfig.json ./examples/project-exporters/vue-nuxt/index.ts"
+    "local:publish": "lerna exec -- npm publish --registry http://localhost:4873",
+    "local:unpublish": "lerna exec -- npm unpublish -f --registry http://localhost:4873"
```
JayaKrishnaNamburu: Just a reminder, should remove this while merging to master, we should always use ``lerna publish`` for that

alexnm: what if we keep them for the local verdaccio setup?

we could use `npm run publish` for the real npm publish and `npm run publish:local` for the verdaccio server

JayaKrishnaNamburu: Yeah, that works too :thumbs_up: but the command should be changed. We should use only ``lerna publish`` even for local because in this way each package is packed and published, but the correct way when we are using lerna is ``lerna publish`` which bumps the version and pushes to npm. This was a temp workaround which i made 😅

alexnm:
@allcontributors[bot] please add @Utwo for code

JayaKrishnaNamburu:
@alexnm i think we should update the coverage collection metric too!

PR #195, opened 2019-04-23
Feature/teleport plugin styled components

### LINKED ISSUE #73: Add plugin for styled-components
Being a pretty popular solution, it would be a shame not to have it as an option for the generated React components

JayaKrishnaNamburu:
Fixes #73 

Just making the PR as WIP, there are few things to do

- Revisit the logic of mapping styles when generating a styled component
- Updating the JSX tags with respect to the generated Styled Components

After solving the both, I will keep this aside and revisit this once the **lerna** migration is done 👍

### DIFF
--- a/src/plugins/teleport-plugin-react-styled-components/index.ts
+++ b/src/plugins/teleport-plugin-react-styled-components/index.ts
@@ -3,6 +3,7 @@ import { ComponentPluginFactory, ComponentPlugin } from '../../typings/generator
 import { generateStyledComponent } from './utils'
 import { ParsedASTNode } from '../../shared/utils/ast-js-utils'
 import { traverseElements, transformDynamicStyles } from '../../shared/utils/uidl-utils'
+import { stringToUpperCamelCase } from '../../shared/utils/string-utils'
 
 interface StyledComponentsConfig {
   componentChunkName: string
@@ -20,12 +21,15 @@ export const createPlugin: ComponentPluginFactory<StyledComponentsConfig> = (con
       return structure
     }
 
+    const jsxNodesLookup = componentChunk.meta.nodesLookup
     const jssStyleMap = {}
 
     traverseElements(node, (element) => {
       const { style, key, elementType } = element
-      const className = `${key}Wrapper`
       if (style) {
+        const root = jsxNodesLookup[key]
+        const className = `${stringToUpperCamelCase(key)}Wrapper`
+        root.openingElement.name.name = className
         jssStyleMap[className] = transformDynamicStyles(
           style,
           (styleValue) =>
--- a/src/plugins/teleport-plugin-react-styled-components/index.ts
+++ b/src/plugins/teleport-plugin-react-styled-components/index.ts
@@ -30,6 +30,7 @@ export const createPlugin: ComponentPluginFactory<StyledComponentsConfig> = (con
         const className = `${stringToUpperCamelCase(key)}Wrapper`
         jssStyleMap[className] = transformDynamicStyles(style, (styleValue) => {
           if (styleValue.content.referenceType === 'prop') {
+            root.openingElement.attributes.push(createJSXSpreadAttribute('props'))
             return `\$\{props => props.${styleValue.content.id}\}`
           }
           throw new Error(
@@ -39,7 +40,6 @@ export const createPlugin: ComponentPluginFactory<StyledComponentsConfig> = (con
           )
         })
         root.openingElement.name.name = className
-        root.openingElement.attributes.push(createJSXSpreadAttribute('props'))
         const code = {
           type: 'js',
           name: className,
--- a/src/plugins/teleport-plugin-react-styled-components/index.ts
+++ b/src/plugins/teleport-plugin-react-styled-components/index.ts
@@ -1,6 +1,8 @@
+import * as t from '@babel/types'
 import { ComponentPluginFactory, ComponentPlugin } from '../../typings/generators'
-import { generateStyledWrapper } from './utils'
-import { traverseElements } from '../../shared/utils/uidl-utils'
+import { generateStyledComponent } from './utils'
+import { ParsedASTNode } from '../../shared/utils/ast-js-utils'
+import { traverseElements, transformDynamicStyles } from '../../shared/utils/uidl-utils'
 
 interface StyledComponentsConfig {
   componentChunkName: string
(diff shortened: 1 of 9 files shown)


src/plugins/teleport-plugin-react-styled-components/index.ts
```diff
@@ -0,0 +1,61 @@
...
+    const componentChunk = chunks.find((chunk) => chunk.name === componentChunkName)
+    if (!componentChunk) {
+      return structure
+    }
+
+    const jssStyleMap = {}
+
+    traverseElements(node, (element) => {
+      const { style, key, elementType } = element
+      const className = `${key}Wrapper`
```
JayaKrishnaNamburu: className should always start with a capital letter, PascalCase for JSX

JayaKrishnaNamburu:
Updated test cases for throwing error scenario, when we try to access state, please refer this issue for more details #200

src/plugins/teleport-plugin-react-styled-components/index.ts
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    const jsxNodesLookup = componentChunk.meta.nodesLookup
+    const jssStyleMap = {}
+
+    traverseElements(node, (element) => {
+      const { style, key, elementType } = element
+      if (style) {
+        const root = jsxNodesLookup[key]
+        const className = `${stringToUpperCamelCase(key)}Wrapper`
```
alexnm: I would drop the Wrapper ending, the key by itself is built to be either the className of certain elements or, in this case, the name of the styled component

src/plugins/teleport-plugin-react-styled-components/index.ts
```diff
@@ -0,0 +1,69 @@
...
+        let needInjectionOfProps: boolean = false
+        jssStyleMap[className] = transformDynamicStyles(style, (styleValue) => {
+          if (styleValue.content.referenceType === 'prop') {
+            needInjectionOfProps = true
```
alexnm: By default we don't need to spread the props, styled-components will send all the props of the elements down to each styled component. Is there a specific reason for adding this behavior by default? if we need to send custom props to the styled component (ex: from state) we should send them explicitly

src/plugins/teleport-plugin-react-styled-components/index.ts
```diff
@@ -0,0 +1,69 @@
...
+            return `\$\{props => props.${styleValue.content.id}\}`
+          }
+          throw new Error(
+            `Error running transformDynamicStyles in reactStyledComponentsPlugin. Unsupported styleValue.content.referenceType value ${
+              styleValue.content.referenceType
+            }`
+          )
+        })
```
alexnm: more of a styling thing, try to leave an empty line after each if block, so the code looks better separated. I'll have to see if prettier/tslint have any specific setting for this

src/plugins/teleport-plugin-react-styled-components/utils.ts
```diff
@@ -0,0 +1,31 @@
...
+      t.identifier(name),
+      t.taggedTemplateExpression(
+        t.memberExpression(t.identifier('styled'), t.identifier(type)),
+        stringAsTemplateLiteral(mapStyles(styles))
+      )
+    ),
+  ])
+}
+
+const mapStyles = (styles: object) => {
```
alexnm: can we something from JSS to compile the style object into the style rules for a single component?

src/plugins/teleport-plugin-react-styled-components/index.ts
JayaKrishnaNamburu: Hey @alexnm check this out --> https://codesandbox.io/s/q7ywk8647j if i am not spreading props and passing them, the component is not rendering the style, initially, i was not doing this, but once i exported the component to sandbox, i observed this behavior

src/plugins/teleport-plugin-react-styled-components/utils.ts
JayaKrishnaNamburu: No @alexnm I tried all possible ways that I am aware of, to do this job using JSS. But ended up writing this method 😅

src/plugins/teleport-plugin-react-styled-components/index.ts
alexnm: ah, I misread the docs. However, does it make sense to pass all props through, or just the ones which are used in the styled component. I think that in most of the cases you need 1 of the props in the styled component. I rarely see the case where all of them should be passed. Hence, I would opt for explicitly passing each prop if it is used in the `style` object

JayaKrishnaNamburu: Sure, then I will pass the props that are being used, which is much needed to control the bloat for the style instead of receiving the whole ``props``, but here is a doubt

For example if the component is using 3-4 attributes from props, then we might end up doing something like this

```<Container height={props.height} width={props.width} />```

What is your opinion we will choose to go with explicit now ?

Comment (alexnm) [reply src/plugins/teleport-plugin-react-styled-components/index.ts]: let's pass { ...props } if at least two values are used in the style object. and let's document this as well in the code with a simple explanation comment

Comment (JayaKrishnaNamburu) [reply src/plugins/teleport-plugin-react-styled-components/index.ts]: Done I will make the changes, if at least two values are being used then I will pass down the props or i will just send the value that is being used 👍

**Comment by JayaKrishnaNamburu** [comment]
Hey @alexnm I was making the suggested changes as we discussed, but here are the few things that are making it trickier

1. When we want to send props explicitly we need to send something like this
~~~
<Contianer color={props.config.color} />
 const Container = styled.div`
  height: ${props => props.height};
`;
~~~
But when we see the resolved UIDL structure the styles comes like this
~~~
{ flexDirection:
   { type: 'dynamic',
     content: { referenceType: 'prop', id: 'direction' } },
  height:
   { type: 'dynamic',
     content: { referenceType: 'prop', id: 'config.height' } },
  alignSelf: { type: 'static', content: 'center' } }
~~~
If we see the ``height`` attribute here when we are mapping the styles our value wil be something like this
~~~
{ type: 'dynamic',
  content: { referenceType: 'prop', id: 'config.height' } }
~~~
which does'nt speak about any thing about the attribute name, so we need to remap and fetch the ``attributeName`` to acheive this.
2. I am not sure if we need to do all this additational calculations and checks just to support the case if the props is refered only once, and we need to consider if the only time that this is being refered should not be a state value, and this is causing around 100ms~ increase in time of the generator.

My opinion is anyways we are passing ``props`` in case of 2, so we keep it same for all cases and brainstorm a better solution like creating dynamic props object and passing it without any nestings, something like that.

**Comment by JayaKrishnaNamburu** [comment]
Hey, @alexnm implemented passing of props explicitly as we discussed and updated test cases.

File: src/plugins/teleport-plugin-react-styled-components/index.ts [hunk 6]
```diff
@@ -0,0 +1,89 @@
...
+
+    traverseElements(node, (element) => {
+      const { style, key, elementType } = element
+      if (style) {
+        const root = jsxNodesLookup[key]
+        const className = `${stringToUpperCamelCase(key)}`
+        let timesReferred: number = 0
+
+        Object.keys(style).map((item) => {
+          // @ts-ignore-next-line
```
JayaKrishnaNamburu: Please check this, it is throwing some ts-lint issue I tried to figure out, but a little confused as we are already referring ``UIDLDynamicReference`` which gives types for ``refereceTypes``

src/plugins/teleport-plugin-react-styled-components/index.ts
```diff
@@ -0,0 +1,83 @@
...
+
+    const jsxNodesLookup = componentChunk.meta.nodesLookup
+    const jssStyleMap = {}
+
+    traverseElements(node, (element) => {
+      const { style, key, elementType } = element
+      if (style) {
+        const root = jsxNodesLookup[key]
+        const className = `${stringToUpperCamelCase(key)}`
+        // @ts-ignore-next-line
```
JayaKrishnaNamburu: Please check this, it is throwing some ts-lint issue I tried to figure out, but a little confused as we are already referring ``UIDLDynamicReference`` which gives types for ``refereceTypes``

PR #196, opened 2019-04-23
feat(lerna): packages structure

alexnm:
moved smaller packages into core and shared


lerna.json
```diff
@@ -2,7 +2,7 @@
   "packages": [
     "packages/*"
   ],
-  "version": "0.0.7",
+  "version": "independent",
```
JayaKrishnaNamburu: Alex, any specific use case  why we are adopting independent instead of maintaining the same version number across all packages. Here are the pros and cons that i felt if we use the default standard

Pros:
- All packages follow the same major version, so when they are being added to project independently we will get to know easily if a specific package with version 2 goes well with our package version 
Cons:
- One major con that i see with this approach is, if we are introducing a breaking change to a specific package and want all the end users to upgrade, then we need to intentionally bump the version number of all packages :grin:

packages/teleport-generator-core/src/validator/index.ts
```diff
@@ -1,6 +1,6 @@
 import Ajv from 'ajv'
-import componentSchema from '@teleporthq/teleport-uidl-definitions/lib/schemas/component.json'
-import projectSchema from '@teleporthq/teleport-uidl-definitions/lib/schemas/project.json'
+import componentSchema from '@teleporthq/teleport-generator-shared/src/uidl-definitions/schemas/component.json'
+import projectSchema from '@teleporthq/teleport-generator-shared/src/uidl-definitions/schemas/project.json'
```
JayaKrishnaNamburu: Hey @alexnm can we adopt the step which I used in the other branch here --> https://github.com/teleporthq/teleport-code-generators/pull/194/files#diff-fa69e88ee0eea386a078ae0237eff9f7R21, since in all the other packages we import things under ``/lib/`` and only in this specific case we directly import from ``src`` to eliminate any confusions while using in future i just wrote a simple copy command to create and expose a ``lib`` instead of ```src`` 😅

packages/teleport-react-component-generator/src/react-component.ts
```diff
@@ -1,29 +1,26 @@
...
-import { createFile } from '@teleporthq/teleport-shared/lib/utils/project-utils'
-import { sanitizeVariableName } from '@teleporthq/teleport-shared/lib/utils/string-utils'
-import { FILE_TYPE } from '@teleporthq/teleport-constants'
+import { createFile } from '@teleporthq/teleport-generator-shared/lib/utils/project-utils'
+import { sanitizeVariableName } from '@teleporthq/teleport-generator-shared/lib/utils/string-utils'
+import { FILE_TYPE } from '@teleporthq/teleport-generator-shared/lib/constants'
 
-import htmlMapping from '@teleporthq/teleport-uidl-definitions/lib/elements-mapping/html-mapping.json'
+import htmlMapping from '@teleporthq/teleport-generator-shared/src/uidl-definitions/elements-mapping/html-mapping.json'
 import reactMapping from './react-mapping.json'
```
JayaKrishnaNamburu: Like here, if we see line 10 and 12 here, one place we import from lib and another place from src

JayaKrishnaNamburu:
PR makes the structure so much cleaner now 🎉

packages/teleport-generator-core/src/validator/index.ts
alexnm: this is just a temporary thing, I wasn't sure where to keep the schemas and the mappings, probably will move them closer to where they are used so there's no need to copy them to the output lib folder. will update on this one

packages/teleport-react-component-generator/src/react-component.ts
alexnm: only .json imports are from src because they are not copied by default by typescript, will fix it

Issue #168, opened 2019-04-09
Showcase project generators in codesandbox using xhr-based sandbox generation


vladnicula:
Our examples and / or docs could use the [XHR Based Generation of CodeSandbox projects](https://codesandbox.io/docs/importing#xhr-request) to generate code sandboxes on the fly for given UIDLs. 

This could be great for testing out project generators and showing examples on the fly. 

Coincidently, the codesandbox api matches the project generator output quite well. 

<img width="368" alt="Screenshot 2019-04-09 at 16 07 52" src="https://user-images.githubusercontent.com/126038/55802808-a7bf8a00-5ae1-11e9-821a-b58cae010552.png">

JayaKrishnaNamburu:
That will an awesome feature @vladnicula ✨

vladnicula:
We'll probably look at this two week from now or so :)

JayaKrishnaNamburu:
I think we can you this for our component generators --> https://github.com/codesandbox/codesandboxer

alexnm:
closing this as we'll resume the discussion on the codesandbox integration and the work will happen most likely in #326

PR #201, opened 2019-04-29
fix(style-flavours): Fixed style flavours w.r.t to their behaviour

### LINKED ISSUE #200: Refactoring and Revisiting our React Styled Flavours
Hi, here are my following observations with the styled flavors that we are currently supporting.

Currently, few style plugins are throwing errors when we are trying to access state variable instead of props in our UIDL.

**UIDL Schema** that is being used

~~~
{
  "$schema": "https://raw.githubusercontent.com/teleporthq/teleport-code-generators/master/src/uidl-definitions/schemas/component.json",
  "name": "StyledComponent",
  "stateDefinitions": {
    "active": {
      "type" : "boolean",
      "defaultValue": false
    }
  },
  "propDefinitions": {
    "direction": {
      "type": "string",
      "defaultValue": "row"
    },
    "config": {
      "type": "object",
      "defaultValue": {
        "height": 32
      }
    }
  },
  "node": {
    "type": "element",
    "content": {
      "elementType": "container",
      "style": {
        "flexDirection": {
          "type": "dynamic",
          "content": {
            "referenceType": "prop",
            "id": "direction"
          }
        },
        "fontSize": {
          "type": "dynamic",
          "content": {
            "referenceType": "state",
            "id": "active"
          }
        },
        "height": {
          "type": "dynamic",
          "content": {
            "referenceType": "prop",
            "id": "config.height"
          }
        },
        "border": "1px solid",
        "alignSelf": {
          "type": "static",
          "content": "center"
        }
      }
    }
  }
}
~~~

### InlineCss
~~~
const StyledComponent = (props) => {
  const [active, setActive] = useState(false)
  return (
    <div
      style={{
        flexDirection: props.direction,
        fontSize: props.active,
        height: props.config.height,
        border: '1px solid',
        alignSelf: 'center',
      }}
    />
  )
}
~~~

``fontSize`` which is a state variable is being mapped with props, need to throw error in this case for ``InlineStyles``

### Css Modules
~~~
const StyledComponent = (props) => {
  const [active, setActive] = useState(false)
  return (
    <div
      style={{
        flexDirection: (props) => props.direction,
        fontSize: (props) => props.active,
        height: (props) => props.config.height,
      }}
      className={styles.container}
    />
  )
}
~~~
``fontSize`` is being mapped to props again here too.

### JSS
~~~
const style = {
  container: {
    flexDirection: (props) => props.direction,
    fontSize: (props) => props.active,
    height: (props) => props.config.height,
    border: '1px solid',
    alignSelf: 'center',
  },
}
~~~

same behavior.

### StyledJSX and StyledComponents
- Styled components (which we have an open PR #73 ) is throwing an error
- StyledJSX is throwing an error 😎 

### TODO

- [ ] Throw errors or do the rectification that is needed.
- [ ] Add Test Cases for the cases where the styles should fail when accessing state values.

JayaKrishnaNamburu:
fixes #200

In this PR, is for discussion and behavior of the style flavors that we currently have and support.
I have shared a few observation in the issue and then I have played around a little bit with the styles support that we currently have and here are the **fixes** that I made to the styles.

PS: Will update the PR shortly in detail with all the observations and fixes that i am working on

#### Things to TODO

- [x] Fix the style plugins of React with respect to their support and behavior.
- [x] Add Rhobust test cases to keep track of all these behaviors, so it is easy to track the changes when the plugin is undergoing any changes in the future.

### Proposal
- Implementing #168 into the development pipeline, so we can refer if the component code that is being generated is a valid one.
- Writing test cases for the generated code using [react-testing-library](https://github.com/testing-library/react-testing-library) in that way we can assess the behavior of the component that is being generated, if the styles are being applied or if the component is rendering or breaking etc..

### DIFF
--- a/src/plugins/teleport-plugin-react-css-modules/index.ts
+++ b/src/plugins/teleport-plugin-react-css-modules/index.ts
@@ -65,16 +65,13 @@ export const createPlugin: ComponentPluginFactory<ReactCSSModulesConfig> = (conf
         if (Object.keys(dynamicStyles).length) {
           const rootStyles = cleanupNestedStyles(dynamicStyles)
 
-          const inlineStyles = transformDynamicStyles(
-            rootStyles,
-            (styleValue) =>
-              new ParsedASTNode(
-                t.arrowFunctionExpression(
-                  [t.identifier('props')],
-                  t.memberExpression(t.identifier('props'), t.identifier(styleValue.content.id))
-                )
-              )
-          )
+          const inlineStyles = transformDynamicStyles(rootStyles, (styleValue) => {
+            const expression =
+              styleValue.content.referenceType === 'state'
+                ? t.identifier(styleValue.content.id)
+                : t.memberExpression(t.identifier('props'), t.identifier(styleValue.content.id))
+            return new ParsedASTNode(expression)
+          })
 
           addJSXTagStyles(root, inlineStyles)
         }
--- a/src/plugins/teleport-plugin-react-inline-styles/index.ts
+++ b/src/plugins/teleport-plugin-react-inline-styles/index.ts
@@ -40,9 +40,11 @@ export const createPlugin: ComponentPluginFactory<InlineStyleConfig> = (config) 
         // Nested styles are ignored
         const rootStyles = cleanupNestedStyles(style)
         const inlineStyles = transformDynamicStyles(rootStyles, (styleValue) => {
-          return new ParsedASTNode(
-            t.memberExpression(t.identifier('props'), t.identifier(styleValue.content.id))
-          )
+          const expression =
+            styleValue.content.referenceType === 'state'
+              ? t.identifier(styleValue.content.id)
+              : t.memberExpression(t.identifier('props'), t.identifier(styleValue.content.id))
+          return new ParsedASTNode(expression)
         })
 
         addJSXTagStyles(jsxASTTag, inlineStyles)
--- a/__tests__/integration/component-assignments/component-with-valid-state-reference.json
+++ b/__tests__/integration/component-assignments/component-with-valid-state-reference.json
@@ -1,0 +1,74 @@
+{
+  "$schema": "https://raw.githubusercontent.com/teleporthq/teleport-code-generators/master/src/uidl-definitions/schemas/component.json",
+  "name": "StyledComponent",
+  "stateDefinitions": {
+    "active": {
+      "type": "string",
+      "defaultValue": "block"
+    }
+  },
+  "propDefinitions": {
+    "direction": {
+      "type": "string",
+      "defaultValue": "row"
+    },
+    "config": {
+      "type": "object",
+      "defaultValue": {
+        "height": 32
+      }
+    }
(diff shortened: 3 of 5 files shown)


alexnm:
@JayaKrishnaNamburu this looks great so far, I do have one comment on the testing ideea. would be really useful to be able to make more rigorous tests with the exported components, but I would wait until we merge the lerna effort to do the setup only once.

JayaKrishnaNamburu:
Sure @alexnm we can look into that once the lerna is done, as it might take a considerable amount of time for us to set up the test infrastructure now and that too we are undergoing some migration 👍 

Here are the changes that I made in this PR.

Inline Styles
- Inline styles accept both props and state, so fixed a bug which is generating ``props.state`` to ``state``

JSS
- JSS supports only props, so throwing an error when the state is being referred.

CSS modules
- This will take both state and props but currently, we are generating arrow expressions but it supports only member expressions when referring props, so correct the output and the test case with respect to it.

JayaKrishnaNamburu:
There is an Edge case which we are currently not considering in all the style plugins, when we are creating a style which refers to props / state, we are not checking if the state is defined in the UIDL, instead we are building the expression, so for example if we are generating a style with ``active`` state but we don't have an active ``state`` defined, This will still generate the code, what is your opinion on this, **should we throw a error/  just generate the code/ warn the user and skip the particular line**.

src/plugins/teleport-plugin-react-css-modules/index.ts
```diff
@@ -65,16 +65,13 @@ export const createPlugin: ComponentPluginFactory<ReactCSSModulesConfig> = (conf
...
-            (styleValue) =>
-              new ParsedASTNode(
-                t.arrowFunctionExpression(
-                  [t.identifier('props')],
-                  t.memberExpression(t.identifier('props'), t.identifier(styleValue.content.id))
-                )
-              )
-          )
+          const inlineStyles = transformDynamicStyles(rootStyles, (styleValue) => {
+            const expression =
```
alexnm: we should do a switch based function here, in case it's not `state` and not `props`

src/plugins/teleport-plugin-react-inline-styles/index.ts
```diff
@@ -40,9 +40,11 @@ export const createPlugin: ComponentPluginFactory<InlineStyleConfig> = (config)
         // Nested styles are ignored
         const rootStyles = cleanupNestedStyles(style)
         const inlineStyles = transformDynamicStyles(rootStyles, (styleValue) => {
-          return new ParsedASTNode(
-            t.memberExpression(t.identifier('props'), t.identifier(styleValue.content.id))
-          )
+          const expression =
+            styleValue.content.referenceType === 'state'
```
alexnm: same as above. I think we can extract a helper function

JayaKrishnaNamburu: Yes I agree, we are doing the same kind of check in a couple of style plugins 👍

PR #216, opened 2019-05-09
fix(react-styled-components): Fixed of adding styled by default

JayaKrishnaNamburu:


JayaKrishnaNamburu:
> can you also write a unit test for the plugin? we just need 2-3 flows to test precisely these scenarios:
> 
> * regular flow when styled-component is added and dependency is also added
> * chunk missing
> * chunk present but style missing (what you just implemented)

Sure, will do that 👍

packages/teleport-generator-shared/src/builders/uidl-builders.ts
```diff
@@ -19,24 +20,30 @@ export const component = (name: string, node: UIDLNode): ComponentUIDL => {
...
+  children?: UIDLNode[],
+  style?: Record<string, UIDLStyleValue>
 ) => {
   return {
     elementType,
     name: elementType,
     attrs,
+    style,
     children,
+    key: elementType,
```
alexnm: why is the `key` configured here? the resolver should generate unique keys for the entire component

alexnm: if it's needed specific for a test to generate the className properly, then it should be added on the spot when setting up that specific test. Otherwise, this function should remain generic as the standard behavior when building a UIDL is to NOT have the `key` attribute and let the resolver handle it

PR #221, opened 2019-05-15
feat(plugin-react-proptypes): Added isRequired attribute for proptypes

### LINKED ISSUE #12: Revisit PropTypes handling
PropTypes are not too flexible at the moment
- [ ] each prop requires a default value to be displayed
- [x] we don't have the ability to set more complex default values (ex: object, array)
- [x] children props don't have a clear definition/standard (also check slots in vue)

### LINKED ISSUE #213: Default props are not being generated in case of only boolean prop is present
When our UIDL have only a single PropDefinition which is a boolean, the ``defaultProps`` and ``propTypes`` are not generating.

UIDL to replicate

~~~
{
  "$schema": "https://raw.githubusercontent.com/teleporthq/teleport-code-generators/master/packages/teleport-generator-core/src/uidl-schemas/component.json",
  "name": "SimpleComponent",
  "propDefinitions": {
    "test": {
      "type": "boolean",
      "defaultValue": true
    }
  },
  "node": {
    "type": "element",
    "content": {
      "elementType": "container",
      "children": ["Hello"]
    }
  }
}
~~~

JayaKrishnaNamburu:
Removed generation of defaultProps if the component doesn't have any defaultValue, for any of the props. And supporting ``isRequired`` attribute for ``propTypes`` definition

fixes #213 fixes #12

### DIFF
--- a/packages/teleport-generator-shared/src/typings/uidl.ts
+++ b/packages/teleport-generator-shared/src/typings/uidl.ts
@@ -34,6 +34,7 @@ export interface ComponentUIDL {
 export interface UIDLPropDefinition {
   type: string
   defaultValue?: string | number | boolean | any[] | object | (() => void)
+  isRequired: boolean
   meta?: Record<string, any>
 }
 
--- a/packages/teleport-generator-shared/src/typings/uidl.ts
+++ b/packages/teleport-generator-shared/src/typings/uidl.ts
@@ -34,7 +34,7 @@ export interface ComponentUIDL {
 export interface UIDLPropDefinition {
   type: string
   defaultValue?: string | number | boolean | any[] | object | (() => void)
-  isRequired: boolean
+  isRequired: false
   meta?: Record<string, any>
 }
 
--- a/packages/teleport-generator-shared/src/typings/uidl.ts
+++ b/packages/teleport-generator-shared/src/typings/uidl.ts
@@ -34,7 +34,7 @@ export interface ComponentUIDL {
 export interface UIDLPropDefinition {
   type: string
   defaultValue?: string | number | boolean | any[] | object | (() => void)
-  isRequired?: false
+  isRequired?: boolean
   meta?: Record<string, any>
 }
 
--- a/packages/teleport-generator-shared/src/typings/uidl.ts
+++ b/packages/teleport-generator-shared/src/typings/uidl.ts
@@ -34,7 +34,7 @@ export interface ComponentUIDL {
 export interface UIDLPropDefinition {
   type: string
   defaultValue?: string | number | boolean | any[] | object | (() => void)
-  isRequired: false
+  isRequired?: false
   meta?: Record<string, any>
 }
 
--- a/packages/teleport-plugin-react-proptypes/__tests__/index.ts
+++ b/packages/teleport-plugin-react-proptypes/__tests__/index.ts
@@ -1,0 +1,132 @@
+import { createPlugin } from '../src'
+import {
+  component,
+  elementNode,
+} from '@teleporthq/teleport-generator-shared/lib/builders/uidl-builders'
+import { ComponentStructure } from '@teleporthq/teleport-generator-shared/lib/typings/generators'
+
+describe('Should add Default Proos and PropTypes to the component Generation', () => {
+  const plugin = createPlugin()
+  const reactChunk = {
+    type: 'js',
+    name: 'react-component',
+    meta: {},
+    content: {},
+    linkAfter: [],
+  }
+  const exportChunk = {
+    type: 'js',
+    name: 'export',
+    content: {},
+    linkAfter: ['react-component'],
+  }
+
+  it('Should throw error when the chunk is supplied', async () => {
+    const uidlSmaple = component('SimpleComponent', elementNode('container'))
+    const structure: ComponentStructure = {
+      uidl: uidlSmaple,
+      chunks: [],
+      dependencies: {},
+    }
+    try {
+      await plugin(structure)
+    } catch (e) {
+      expect(e.message).toContain('React component chunk with name')
+    }
+  })
+
+  it('Should generate chunks, defaultProps and propTypes', async () => {
+    const props = {
+      test: {
+        type: 'boolean',
+        defualtValue: 'true',
+      },
+      name: {
+        type: 'string',
+        defaultValue: 'Teleport',
+      },
+    }
+    // @ts-ignore-next-line
+    const uidlSmaple = component('SimpleComponent', elementNode('container'), props)
+    const structure: ComponentStructure = {
+      uidl: uidlSmaple,
+      chunks: [reactChunk, exportChunk],
+      dependencies: {},
+    }
+    const result = await plugin(structure)
+
+    const defaultProps = result.chunks.filter(
+      (chunk) => chunk.name === 'react-component-default-props'
+    )
+    const propTypes = result.chunks.filter(
+      (chunk) => chunk.name === 'react-component-types-of-props'
+    )
+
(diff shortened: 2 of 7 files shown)


packages/teleport-generator-shared/src/typings/uidl.ts
```diff
@@ -34,6 +34,7 @@ export interface ComponentUIDL {
 export interface UIDLPropDefinition {
   type: string
   defaultValue?: string | number | boolean | any[] | object | (() => void)
+  isRequired: boolean
```
alexnm: we could set a default value of false

packages/teleport-plugin-react-proptypes/src/index.ts
```diff
@@ -37,10 +37,16 @@ export const createPlugin: ComponentPluginFactory<ReactJSPropTypesConfig> = (con
     }
 
     // TODO used the name from the mappings of the component, not from the UIDL
-    const defaultPropsAst = buildDefaultPropsAst(name, uidl.propDefinitions)
+    let hasDefaultProps: boolean = false
+    Object.keys(uidl.propDefinitions).forEach((prop) => {
```
alexnm: you can use Array.prototype.some here to avoid doing a condition - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some

packages/teleport-plugin-react-proptypes/src/index.ts
```diff
@@ -65,7 +75,7 @@ export const createPlugin: ComponentPluginFactory<ReactJSPropTypesConfig> = (con
     })
 
     // push export of component after declarations of types
-    exportChunk.linkAfter.push(typesOfPropsChunkName, defaultPropsChunkName)
+    exportChunk.linkAfter.push(typesOfPropsChunkName)
```
alexnm: will this invert the order? default props coming before types? I would keep the original order as this is how components usually are defined, first propTypes, then defaultProps

packages/teleport-plugin-react-proptypes/src/utils.ts
```diff
@@ -17,10 +17,8 @@ export const buildDefaultPropsAst = (
   const defaultValuesSearch = Object.keys(propDefinitions).reduce(
     (acc: any, key) => {
       const { defaultValue } = propDefinitions[key]
-      if (defaultValue) {
-        acc.values[key] = defaultValue
-        acc.count++
-      }
+      acc.values[key] = defaultValue
```
alexnm: what if this current prop does not have a defaultValue?

packages/teleport-plugin-react-proptypes/src/index.ts
JayaKrishnaNamburu: I think the order is fine with the UIDL's I tested, but let me recheck few more 👍

packages/teleport-plugin-vue-base-component/src/utils.ts
```diff
@@ -318,7 +318,12 @@ const createVuePropsDefinition = (uidlPropDefinitions: Record<string, UIDLPropDe
         )
     }
 
-    acc[name] = defaultValue ? { type: mappedType, default: defaultValue } : mappedType
+    acc[name] =
+      defaultValue && typeof defaultValue !== undefined
```
alexnm: typeof defaultValue !== undefined should be enough

packages/teleport-plugin-react-proptypes/src/index.ts
```diff
@@ -37,10 +37,15 @@ export const createPlugin: ComponentPluginFactory<ReactJSPropTypesConfig> = (con
     }
 
     // TODO used the name from the mappings of the component, not from the UIDL
-    const defaultPropsAst = buildDefaultPropsAst(name, uidl.propDefinitions)
+    const hasDefaultProps = Object.keys(uidl.propDefinitions).some(
+      (prop) =>
+        uidl.propDefinitions[prop].defaultValue &&
```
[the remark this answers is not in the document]
JayaKrishnaNamburu: No, sometimes if we won't initialize ``defaultValie`` at all we will run into this issue. At first i did not add that but when writing test cases noticed this behavior.

packages/teleport-plugin-react-proptypes/__tests__/index.ts
```diff
@@ -0,0 +1,132 @@
...
+    const props = {
+      test: {
+        type: 'boolean',
+      },
+      name: {
+        type: 'string',
+      },
+    }
+    // @ts-ignore-next-line
+    const uidlSmaple = component('SimpleComponent', elementNode('container'), props)
```
JayaKrishnaNamburu: And @alexnm I have an issue with the typings again here 😅 , can you check once.

packages/teleport-plugin-react-proptypes/__tests__/index.ts
```diff
@@ -0,0 +1,132 @@
...
+      (chunk) => chunk.name === 'react-component-types-of-props'
+    )
+
+    expect(defaultProps.length).toEqual(0)
+    expect(propTypes.length).toEqual(1)
+    expect(propTypes[0].type).toBe('js')
+  })
+
+  it('Should generate chunks after specifying required to props', async () => {
+    const props = {
```
JayaKrishnaNamburu: Alex you can check the ``typeofdefaultValue !== undefined`` case here

packages/teleport-plugin-react-proptypes/src/index.ts
JayaKrishnaNamburu: Only in this case we need to check becasue of the this behaviour
~~~
typeof undefined !== undefined // true
~~~

This returns ``true`` which in returns marks the flag as ``hasDefaultProps`` to ``true`` and an empty tag gets generated.

alexnm: ah, I missed this, it should be `typeof uidl.propDefinitions[prop].defaultValue !== 'undefined'` and that covers all cases

JayaKrishnaNamburu: Done 👍

packages/teleport-plugin-react-proptypes/src/utils.ts
```diff
@@ -17,7 +17,7 @@ export const buildDefaultPropsAst = (
   const defaultValuesSearch = Object.keys(propDefinitions).reduce(
     (acc: any, key) => {
       const { defaultValue } = propDefinitions[key]
-      if (defaultValue) {
+      if (typeof defaultValue !== undefined) {
```
alexnm: 'undefined'

packages/teleport-plugin-vue-base-component/src/utils.ts
```diff
@@ -318,7 +318,10 @@ const createVuePropsDefinition = (uidlPropDefinitions: Record<string, UIDLPropDe
         )
     }
 
-    acc[name] = defaultValue ? { type: mappedType, default: defaultValue } : mappedType
+    acc[name] =
+      typeof defaultValue !== undefined ? { type: mappedType, default: defaultValue } : mappedType
```
alexnm: 'undefined' here as well

packages/teleport-plugin-vue-base-component/src/utils.ts
```diff
@@ -318,7 +318,10 @@ const createVuePropsDefinition = (uidlPropDefinitions: Record<string, UIDLPropDe
...
+    acc[name] = isRequired ? { ...{ required: isRequired }, ...acc[name] } : acc[name]
```
alexnm: this can be translated into { required: isRequired, ...acc[name] }

PR #222, opened 2019-05-15
feat(postprocessor-starptech-html): A new postprocessor for html file

JayaKrishnaNamburu:
A new postprocessor for HTML files using starptech prettyhtml.

I am not sure, how useful this can be but it is very lightweight than the prettier that we are using currently.

### DIFF
--- a/packages/teleport-postprocessor-starptech-html/README.md
+++ b/packages/teleport-postprocessor-starptech-html/README.md
@@ -1,0 +1,11 @@
+# `@teleporthq/teleport-postprocessor-starptech-html`
+
+> TODO: description
+
+## Usage
+
+```
+const teleportPostprocessorStarptechHtml = require('@teleporthq/teleport-postprocessor-starptech-html');
+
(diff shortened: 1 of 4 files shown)


JayaKrishnaNamburu:
This needs a little work, it is formatting vue templates but when I try to run performance test cases using this, tests are breaking. @alexnm what you say should i need to continue or close some issues from 0.7 and revisit this again once we are stable.

JayaKrishnaNamburu:
> forgot about one thing, can you please write a quick suite of unit tests for the post-processor? similar with what we have for the prettier one

yeah sure, I will update this. Right now there is only on bug which is causing this test case to break, looking into that as well.

JayaKrishnaNamburu:
Closing this, as @starptech-prettyhtml uses @starptech-pretty-formatter which in terms has prettier in this dependency, which is resulting in big bundle size and we have a straight forwards prettier based formatter already as a postprocessor plugin.

PR #223, opened 2019-05-15
Fix/#175 file names

### LINKED ISSUE #175: Component names are not same as specified in UIDL
After this #168, I tried exporting a project from CLI too code sandbox using ``codesandbox cli`` and a sandbox is created here --> https://codesandbox.io/s/8zz2kr8118 and then here is the bug the component name is not the same as that is specified in the UIDL is 
~~~
        {
            "value": "contact-us",
            "meta": {
              "path": "/here-we-are",
              "componentName": "Us"
            }
          }
~~~

And the import statement is being generated properly
~~~
import Us from './pages/contact-us'
~~~

But the file name that is being generated is something like this 
~~~
contactus.css
contactus.js
~~~

alexnm:
Update to the string conversion functions + tests. We will go with a standard from now on:
- all component names when used in JS will be camelCase or UpperCamelCase
- all file names will be dashCase (including the import statements for local dependencies)

Fix #175

### DIFF
--- a/examples/uidl-samples/project.json
+++ b/examples/uidl-samples/project.json
@@ -244,7 +244,7 @@
       }
     },
     "OneComponent": {
-      "name": "New UIDL",
+      "name": "OneComponent",
       "propDefinitions": {
         "title": {
           "type": "string",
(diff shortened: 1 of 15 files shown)


alexnm:
> 175

I will check and come back to this

JayaKrishnaNamburu:
Changed the title as this goes into the changelog.

Issue #215, opened 2019-05-08
Validator irregularities/bugs


alexnm:
I noticed two bugs caused by the recent changes in the UIDL structure:

1. When adding a new key inside any `type`/`content` node, there's no validation error (additional key). Look for `asda` key below:

![Screenshot 2019-05-08 at 16 34 55](https://user-images.githubusercontent.com/9945366/57379262-4279ba00-71af-11e9-9a4a-0b3703abbc03.png)

2. When adding a new key inside an `element` node, the validation message is not clear, the 6th line only tells the actual problem (additional key). This is a bit harder to solve because of the new schema, where the node inside `content` can be of multiple types.

![Screenshot 2019-05-08 at 16 34 37](https://user-images.githubusercontent.com/9945366/57379368-79e86680-71af-11e9-8646-64a0af68b66a.png)

PR #224, opened 2019-05-16
Feature/validator extension + bug fix 

anamariaoros:
This MR tackles both #41 and #215


packages/teleport-generator-core/src/validator/index.ts
```diff
@@ -38,6 +42,66 @@ export default class Validator {
 
     return { valid: true, errorMsg: '' }
   }
+
+  public validateComponentContent(input: any): ValidationResult {
+    const verifyProps = checkContent(input, 'prop')
```
alexnm: can you check both prop and state into a single pass? might be a micro-optimization, but when we validate a project of 1000 components in theory, it would mean a lot of passes through the UIDL with traverse

packages/teleport-generator-core/src/validator/index.ts
```diff
@@ -38,6 +42,66 @@ export default class Validator {
...
+        const errorMsg = `"${
+          node.content.id
+        }" is used as ${type} but not defined. Please add it in ${definitions}`
+        errors.push(errorMsg)
+      }
+      usedKeys.push(node.content.id)
+    }
+  })
+
+  const diffs = definedKeys.filter((x) => !usedKeys.includes(x))
```
alexnm: you can pipe the `filter` and the `map` as the map will be skipped in case no values are left after the filter

packages/teleport-generator-core/src/validator/index.ts
```diff
@@ -38,6 +42,66 @@ export default class Validator {
...
+        errorMsg: `\nUIDL Content Validation Error. Please check the following: \n${errors}`,
+      }
+    }
+
+    return { valid: true, errorMsg: '' }
+  }
+}
+
+const checkForDuplicates = (input: any) => {
+  const props = Object.keys(input.propDefinitions || [])
```
alexnm: the default value for `input.propDefinitions` should be an empty object, not empty array

packages/teleport-generator-core/src/validator/index.ts
alexnm: also, I would give the functions a more suggestive name, "checkDynamicDefinitions"?

packages/teleport-generator-core/src/validator/index.ts
```diff
@@ -38,6 +42,66 @@ export default class Validator {
 
     return { valid: true, errorMsg: '' }
   }
+
+  public validateComponentContent(input: any): ValidationResult {
+    const verifyProps = checkContent(input, 'prop')
+    const verifyState = checkContent(input, 'state')
+
+    checkForDuplicates(input)
```
alexnm: same as above about the naming. once we will test more than just props and state here, duplicates will not be suggestive enough

packages/teleport-generator-core/src/validator/utils.ts
```diff
@@ -0,0 +1,186 @@
...
+
+  return { errors }
+}
+
+// The "root" node should contain only elements of type "conditional"
+export const checkRootComponent = (input: ProjectUIDL) => {
+  const errors = []
+  const routeNaming = []
+  const rootNode = input.root.node.content as UIDLElement
+  rootNode.children.map((child) => {
```
alexnm: it's better to use a forEach when you're not returning something in the callback function

packages/teleport-generator-core/src/validator/utils.ts
```diff
@@ -0,0 +1,186 @@
...
+//      }
+//    }
+export const checkComponentNaming = (input: ProjectUIDL) => {
+  const errors = []
+  const namesUsed = Object.keys(input.components)
+
+  namesUsed
+    .filter((name) => input.components[name].name !== name)
+    .map((diff) => {
+      const errorMsg = `\nThe following dependencies have different name than their key: ${diff}`
```
alexnm: shouldn't the message be displayed only once and the the list of components? comma separated?

alexnm: also, I would name them "components", not "dependencies", not to be confused with the dependencies inside elementNodes

packages/teleport-generator-core/src/validator/index.ts
```diff
@@ -62,14 +62,14 @@ export default class Validator {
...
+    const verifyRoutes = utils.checkRouteDefinition(input)
     let errors = verifyRoutes.errors
 
     if (verifyRoutes.errors.length === 0) {
-      const verifyRootComponent = checkRootComponent(input)
+      const verifyRootComponent = utils.checkRootComponent(input)
 
-      const verifyComponentNaming = checkComponentNaming(input)
-      const verifyComponentExistence = checkComponentExistence(input)
+      const verifyComponentNaming = utils.checkComponentNaming(input)
```
alexnm: is there any benefit from having these functions return an array of errors inside an object? can't the functions return just the array? so when it's empty, nothing is displayed from that area

packages/teleport-generator-core/src/validator/index.ts
```diff
@@ -62,14 +62,14 @@ export default class Validator {
...
+      const verifyComponentExistence = utils.checkComponentExistence(input)
```
alexnm: once all the functions returns an array of errors, you can spread them together to form a single array of error messages

packages/teleport-generator-core/src/validator/utils.ts
```diff
@@ -0,0 +1,186 @@
...
+} from '@teleporthq/teleport-generator-shared/lib/typings/uidl'
+
+// Prop definitions and state definitions should have different keys
+export const checkForDuplicateDefinitions = (input: ComponentUIDL) => {
+  const props = Object.keys(input.propDefinitions || {})
+  const states = Object.keys(input.stateDefinitions || {})
+
+  props
+    .filter((x) => states.includes(x))
+    .map((duplicate) =>
```
alexnm: forEach if you are not returning from map

packages/teleport-generator-core/src/validator/utils.ts
```diff
@@ -0,0 +1,186 @@
...
+  const definedKeys = Object.keys(input.propDefinitions || {}).concat(
+    Object.keys(input.stateDefinitions || {})
+  )
+  const usedKeys = []
+  const errors = []
+
+  traverseNodes(input.node, (node) => {
+    if (
+      node.type === 'dynamic' &&
+      (node.content.referenceType === 'prop' || node.content.referenceType === 'state')
```
alexnm: I think you need to check separately here, if refType is prop, you should explicitly check the key exists in `propDefinitions`. similarly, when the refType is state, you should check in `stateDefinitions`

packages/teleport-generator-core/src/validator/utils.ts
```diff
@@ -0,0 +1,186 @@
...
+    errors.push(errorMsg)
+  }
+  return { errors }
+}
+
+// All referenced components inside of the projectUIDL should be defined
+// in the component section
+export const checkComponentExistence = (input: ProjectUIDL) => {
+  const errors = []
+  const dependencies = Object.keys(input.components)
```
alexnm: again, components would be more suggestive here, not to be confused with the other dependency node

PR #226, opened 2019-05-17
Feature/#197 project packer

ionutpasca:


JayaKrishnaNamburu:
Hi @ionutpasca looks awesome, yet to check totally. Can you please remove the package-lock file and re add the dependency using yarn and push the yarn.lock file. Since we switched from nom to yarn with the adoption of lerna.

ionutpasca:
> Hi @ionutpasca looks awesome, yet to check totally. Can you please remove the package-lock file and re add the dependency using yarn and push the yarn.lock file. Since we switched from nom to yarn with the adoption of lerna.

Hello @JayaKrishnaNamburu , thank you for the feedback. Ofcourse, I will. There will be a few more changes and I needed some feedback before the final version of this feature was done. I'll also add a few more tests so that the coverage will be as high as possible, the github integration performance will be improved as well and if there is some duplicated code I'll extract it into some separate files.
If you have additional feedback, I'll be glad to hear your opinion.

package.json
```diff
@@ -45,6 +45,7 @@
     "lerna-changelog": "^0.8.2",
     "pretty-quick": "^1.10.0",
     "ts-jest": "^24.0.2",
+    "ts-node": "^8.1.0",
```
alexnm: why is ts-node needed in the root package.json?

packages/teleport-generator-shared/src/typings/generators.ts
```diff
@@ -137,13 +137,13 @@ export interface ProjectGeneratorOptions {
...
 }
 
-export type ProjectGeneratorFunction = (
-  uidl: ProjectUIDL,
-  options?: ProjectGeneratorOptions
-) => Promise<{
-  outputFolder: GeneratedFolder
-  assetsPath?: string
-}>
+// export type ProjectGeneratorFunction = (
```
alexnm: you can remove this interface, it is only used by the project exporters in the example folder and should not be here anyway

packages/teleport-generator-shared/src/typings/generators.ts
```diff
@@ -187,11 +187,71 @@ export interface ProjectGeneratorOutput {
...
+
+/**
+ * Interfaces used in the publishers
+ */
+export type PublisherFactory<T, U> = (configuration?: Partial<T & PublisherFactoryParams>) => U
+
+export interface Publisher<T, U> {
+  publish: (options?: T) => Promise<PublisherResponse<U>>
+  getProject: () => GeneratedFolder | undefined
+  setProject: (project: GeneratedFolder) => unknown
```
alexnm: can you use void instead of undefined or unknown?

packages/teleport-generator-shared/src/typings/generators.ts
```diff
@@ -187,11 +187,71 @@ export interface ProjectGeneratorOutput {
...
+  data: string
+  name: string
+  type: string
+}
+
+export interface TemplateDefinition {
+  templateFolder?: GeneratedFolder
+  remote: RemoteTemplateDefinition
+  meta?: {
+    componentsPath?: string[]
```
alexnm: would this work better as a Record<string, string[]> structure? this way you don't have to hardcode here all the possible values

packages/teleport-generator-shared/src/utils/project-utils.ts
```diff
@@ -248,14 +249,144 @@ export const createFile = (name: string, fileType: string, content: string): Gen
...
 ): GeneratedFolder => {
+  let folder = findFolderByPath(rootFolder, path)
+
+  if (!folder) {
+    const { updatedRootFolder, createdFolder } = createFolderByPath(rootFolder, path)
+    rootFolder = updatedRootFolder
+    folder = createdFolder
+  }
+
+  folder.files = folder.files.concat(files)
```
alexnm: you can do folder.files.push(files)

packages/teleport-generator-shared/src/utils/project-utils.ts
```diff
@@ -248,14 +249,144 @@ export const createFile = (name: string, fileType: string, content: string): Gen
...
+
+const removeCommonStartingPointsFromPaths = (
+  componentsPath: string[],
+  pagesPath: string[]
+): { componentsPath: string[]; pagesPath: string[] } => {
+  const componentsPathLength =
+    componentsPath.length > pagesPath.length ? componentsPath.length : pagesPath.length
+
+  let i = 0
+  let isEqual = true
```
alexnm: i and isEqual are a bit too generic in this instance, you need to look at the code in depth to understand what's their meaning

packages/teleport-generator-shared/src/utils/project-utils.ts
```diff
@@ -248,14 +249,144 @@ export const createFile = (name: string, fileType: string, content: string): Gen
...
+  dependencyPrefix += componentsPath
+    .map((component) => {
+      return `${component}/`
+    })
+    .join('')
+
+  return dependencyPrefix
+}
+
+const removeCommonStartingPointsFromPaths = (
```
alexnm: can this function work on any two concepts / folders? or is it only for pages and components?

packages/teleport-project-generator-react-basic/src/utils.ts
```diff
@@ -1,18 +1,37 @@
...
 } from '@teleporthq/teleport-generator-shared/lib/typings/generators'
+import { injectFilesToPath } from '@teleporthq/teleport-generator-shared/lib/utils/project-utils'
+import {
+  DEFAULT_SRC_FILES_PATH,
+  DEFAULT_COMPONENT_FILES_PATH,
+  DEFAULT_PAGE_FILES_PATH,
+  DEFAULT_STATIC_FILES_PATH,
+} from './constants'
 
 export const buildFolderStructure = (
```
alexnm: can this function be reused for all project generators? I think now we're still relying on the parent-child relationship between the src / pages / components folders but it would be nice to try a generic function based on the template / default structure. let's talk about this maybe

packages/teleport-project-packer-playground/package.json
```diff
@@ -0,0 +1,36 @@
+{
+  "name": "@teleporthq/teleport-project-packer-playground",
+  "version": "0.6.0",
+  "description": "Teleport playground packer",
```
alexnm: Write a bit more about the purpose of the playground-packer

packages/teleport-project-packer/package.json
```diff
@@ -0,0 +1,35 @@
...
+  },
+  "publishConfig": {
+    "access": "public"
+  },
+  "scripts": {
+    "build": "tsc"
+  },
+  "dependencies": {
+    "@teleporthq/teleport-generator-shared": "^0.6.0",
+    "@teleporthq/teleport-project-generator-react-basic": "^0.6.0",
```
alexnm: is the generic packer using the generators and the publishers as direct references? or is this just temporary until we figure out the interface?

package.json
ionutpasca: I added the ts-node dependency for debugging purposes. As the package is written in typescript, in order to debug it in vscode and not build it first, you need to run any lerna package you want to debug using the ts-node module.

packages/teleport-generator-shared/src/typings/generators.ts
ionutpasca: Record<string, string[]> would work here, but you won't have autocomplete when defining the template metadata. Let's have a detailed talk about this.

packages/teleport-generator-shared/src/utils/project-utils.ts
ionutpasca: Well, actually I can't. The **folder.files** entity is an array of **GeneratedFiles** and the **files** that I'm trying to inject into the already existing array is also of type **GeneratedFiles[]**. If I would use the push method I would get a matrix (array of array) not a flat structure, like I need.

packages/teleport-generator-shared/src/utils/project-utils.ts
ionutpasca: Renamed:
i -> componentIndex
isEqual -> elementsFromIndexAreEqual

packages/teleport-project-packer/package.json
ionutpasca: It's only temporary. I will remove the dependencies from the package.json file

packages/teleport-generator-shared/src/utils/project-utils.ts
ionutpasca: I've made this function more generic. It now accepts an array of path arrays and removes the common starting points from all of them

PR #231, opened 2019-05-23
fix(vue-base-component): Support for nested conditionals in Vue

### LINKED ISSUE #209: Support nested conditionals
Nested conditionals work in React by adding more conditions in the JSX expression ```{ condition1 && condition2 && content }```

But in Vue, it seems that only the first conditional is taken into consideration.

This is an example UIDL I was testing with:

```
{
  "type": "conditional",
  "content": {
    "reference": {
      "type": "dynamic",
      "content": {
        "referenceType": "state",
        "id": "isVisible"
      }
    },
    "value": true,
    "node": {
      "type": "conditional",
      "content": {
        "node": {
          "type": "element",
          "content": {
            "elementType": "text",
            "children": [
              {
                "type": "static",
                "content": "Now you see me!"
              }
            ]
          }
        },
       "reference": {
          "type": "dynamic",
          "content": {
            "referenceType": "prop",
            "id": "title"
          }
        },
        "value": "asdfasf"
      }
    }
  }
}
```

JayaKrishnaNamburu:
fixes #209


packages/teleport-plugin-vue-base-component/src/utils.ts
```diff
@@ -145,19 +145,30 @@ export const generateConditionalNode = (
...
   const { reference, value } = node.content
   const conditionalKey = reference.content.id
+  let conditionalStatement: string = createConditionalStatement(
+    conditionalKey,
+    generateConditionalExpression(value, node.content.condition)
+  )
 
-  // 'v-if' needs to be added on a tag, so in case of a text node we wrap it with
-  // a 'span' which is the less intrusive of all
+  Object.values(node).forEach((child) => {
```
alexnm: isn't this just going one level down? what about the case when 3 conditionals are nested? I know it's not a common case, but we need to make sure we support everything that the UIDL can represent

packages/teleport-plugin-vue-base-component/src/utils.ts
```diff
@@ -145,19 +145,30 @@ export const generateConditionalNode = (
...
+    if (typeof child === 'object') {
```
alexnm: also, you can assume node.content.node is the next UIDLNode inside, since the input is of type UIDLConditionalNode.

packages/teleport-plugin-vue-base-component/src/utils.ts
```diff
@@ -168,6 +166,38 @@ export const generateConditionalNode = (
   return conditionalTag
 }
 
+const generateConditionalStatement = (node: UIDLNode, statement: string) => {
+  Object.values(node).forEach((child: any) => {
```
JayaKrishnaNamburu: Hey @alexnm  can you check the typing here

PR #239, opened 2019-05-29
fix(vue-base): Appending keys for static data object identifier

### LINKED ISSUE #210: Static data source can overlap prop or state
When a static dataSource is used for a repeat node, the content (an array of values) is added on the `dataObject` (`teleport-plugin-vue-base-component` in `utils.ts`, line 124).

If no meta values are specified, the a default `items` name is used. This can easily overlap with props or state, as visible in the [repl](https://repl.teleporthq.io/) on the `NewComponent` example.

One idea is to pass the parent node inside `generateRepeatNode` and use its unique key to generate the name of the data property (ex: listItems, list1Items, etc.)

JayaKrishnaNamburu:
fixes #210

### DIFF
--- a/packages/teleport-plugin-vue-base-component/src/utils.ts
+++ b/packages/teleport-plugin-vue-base-component/src/utils.ts
@@ -20,6 +20,7 @@ import {
   UIDLRepeatNode,
   UIDLConditionalNode,
   UIDLSlotNode,
+  UIDLElement,
 } from '@teleporthq/teleport-generator-shared/lib/typings/uidl'
 import {
   NodeSyntaxGenerator,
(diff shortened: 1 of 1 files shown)


packages/teleport-plugin-vue-base-component/src/utils.ts
```diff
@@ -117,7 +118,9 @@ export const generateRepeatNode = (
   const { dataSource, node: repeatContent, meta = {} } = node.content
   const repeatContentTag = generateNodeSyntax(repeatContent, accumulators)
 
-  let dataObjectIdentifier = meta.dataSourceIdentifier || `items`
+  const { key } = node.content.node.content as UIDLElement
```
alexnm: the child node is not necessarily an UIDLElement here, it can be any type of UIDLNode

JayaKrishnaNamburu: I tried having it as UIDLNode, but typescript is throwing error for that, saying --> Key does not exist on UIDLNode

alexnm: no sure if this is a solution, as we have no certainty that there's an element with a key there.

JayaKrishnaNamburu: All the nodes will have keys right? (I am not sure about this)

alexnm: only type element have keys, they are built based on the `name`, which again is only inside the content of element nodes.

alexnm:
closing this since @ionutpasca is implementing a generic solution for the static data sources problem

Issue #243, opened 2019-06-04
Mapping group to <template> in Vue


alexnm:
The `react` mapping handles the `group` elementType as:

```json
"group": {
  "elementType": "Fragment",
  "dependency": {
    "type": "library",
    "path": "react",
    "meta": {
      "namedImport": true
    }
  }
}
```

Similarly, the `vue` mapping needs to have a working solution. The first idea was to map that to a <template> tag, but I would allocate some time for testing and ensuring this solution is ok
```json
"group": {
  "elementType": "template"
}
```

alexnm:
Apparently Vue needs a single root node in the <template> and there's no alternative like React.Fragment. For now, I will switch the vue mapping to map the `group` element to a `div`

PR #253, opened 2019-06-06
Fix/project.json

JayaKrishnaNamburu:
The `vue-mapping` currently maps `group` to `template`, so the current UIDL is resulting in creating one more `template` tag apart from the parent. Which are not being rendered when we try to execut. So changed them to `container` so it generates `div` instead.

Here is the result that is generated with the current UIDL --> https://github.com/JayaKrishnaNamburu/teleport-github-packer-test/blob/master/components/navbar.vue#L2

### DIFF
--- a/examples/uidl-samples/project.json
+++ b/examples/uidl-samples/project.json
@@ -4,7 +4,7 @@
   "globals": {
     "settings": {
       "language": "en",
-      "title": "UIDL v0.6 Project"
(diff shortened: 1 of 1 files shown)


alexnm:
there's an issue about this #243

JayaKrishnaNamburu:
> there's an issue about this #243

Oh, i kissed it out. Will check of anything is possible in vue which behaves like react fragments 👍

PR #256, opened 2019-06-06
Refactor/project packer interfaces

alexnm:


packages/teleport-project-packer-playground/src/index.ts
```diff
@@ -152,7 +149,23 @@ export const createPlaygroundPacker = (params: PackerFactoryParams = {}) => {
     packer.setAssets(projectAssets)
     packer.setGeneratorFunction(projectGenerator.generateProject)
     packer.setPublisher(projectPublisher)
-    packer.setTemplate(projectTemplate)
+
+    const remoteGithubTemplate =
+      projectTemplates[packTechnology.type] || projectTemplates[TEMPLATES.REACT_NEXT]
+
+    if (packParams.template) {
```
alexnm: @ionutpasca please review this flow here

packages/teleport-project-packer/src/index.ts
```diff
@@ -49,37 +47,24 @@ export const createProjectPacker: PackerFactory = (params: PackerFactoryParams =
...
+  ): Promise<GeneratedFolder> => {
     try {
-      template.templateFolder = await fetchTemplate(template.remote)
-      templateLoaded = true
-      return { success: true, payload: template.templateFolder }
+      template = await fetchTemplate(remoteDefinition)
+      return template // TODO: is this useful?
     } catch (err) {
-      return { success: false, payload: err }
+      throw err // TODO: error handling here?
```
ionutpasca: Maybe we should only return a response similar to this

```js
{
  success: false
  payload: err
}
```

packages/teleport-project-packer/src/index.ts
```diff
@@ -49,37 +47,24 @@ export const createProjectPacker: PackerFactory = (params: PackerFactoryParams =
...
-    }
-
+    remoteDefinition: RemoteTemplateDefinition
+  ): Promise<GeneratedFolder> => {
     try {
-      template.templateFolder = await fetchTemplate(template.remote)
-      templateLoaded = true
-      return { success: true, payload: template.templateFolder }
+      template = await fetchTemplate(remoteDefinition)
+      return template // TODO: is this useful?
```
ionutpasca: There might be cases in which, after the user loads the template, wants to parse it in some way. So I think that returning the template after we are loading it is a good thing. Maybe just:

```js
return await fetchTemplate(remoteDefinition)
```

alexnm: but then the template inside the packer closure would not be updated. we would have to call setTemplate after loadTemplate

PR #266, opened 2019-06-11
fix(styled-jsx/index): Injecting style in first child if root is non element

### LINKED ISSUE #245: styled-jsx should be injected in the first element node
Currently the styled-jsx plugin assumes the top level node inside the component is of type `element`, but as of #182, we can have any node type as the root node. The plugin should traverse the tree until the first element is found and place the <style> tag there

JayaKrishnaNamburu:
fixes #245

### DIFF
--- a/packages/teleport-plugin-react-styled-jsx/src/index.ts
+++ b/packages/teleport-plugin-react-styled-jsx/src/index.ts
@@ -7,6 +7,7 @@ import { camelCaseToDashCase } from '@teleporthq/teleport-shared/lib/utils/strin
 import {
   transformDynamicStyles,
   traverseElements,
+  traverseUIDLForRootElmNode,
 } from '@teleporthq/teleport-shared/lib/utils/uidl-utils'
 import { createCSSClassFromStringMap } from '@teleporthq/teleport-shared/lib/utils/jss-utils'
 import { ComponentPluginFactory, ComponentPlugin, UIDLElement } from '@teleporthq/teleport-types'
@@ -20,7 +21,6 @@ export const createPlugin: ComponentPluginFactory<StyledJSXConfig> = (config) =>
 
   const reactStyledJSXChunkPlugin: ComponentPlugin = async (structure) => {
     const { uidl, chunks } = structure
-
     const { node } = uidl
 
     const componentChunk = chunks.find((chunk) => chunk.name === componentChunkName)
@@ -62,10 +62,14 @@ export const createPlugin: ComponentPluginFactory<StyledJSXConfig> = (config) =>
     // We have the ability to insert the tag into the existig JSX structure, or do something else with it.
     // Here we take the JSX <style> tag and we insert it as the last child of the JSX structure
     // inside the React Component
+    let rootJSXNode
     const rootNodeKey = (node.content as UIDLElement).key
-    const rootJSXNode = rootNodeKey
-      ? jsxNodesLookup[rootNodeKey]
-      : jsxNodesLookup[Object.keys(jsxNodesLookup)[0]]
+    if (rootNodeKey) {
+      rootJSXNode = jsxNodesLookup[rootNodeKey]
+    } else {
+      const elmRootNode = traverseUIDLForRootElmNode(uidl.node)
+      rootJSXNode = jsxNodesLookup[elmRootNode.content.key]
+    }
 
     rootJSXNode.children.push(jsxASTNodeReference)
     return structure
--- a/packages/teleport-plugin-react-styled-jsx/src/index.ts
+++ b/packages/teleport-plugin-react-styled-jsx/src/index.ts
@@ -7,10 +7,10 @@ import { camelCaseToDashCase } from '@teleporthq/teleport-shared/lib/utils/strin
 import {
   transformDynamicStyles,
   traverseElements,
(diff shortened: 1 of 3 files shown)


JayaKrishnaNamburu:
Any idea why the tests are breaking, in my local all tests are passing
Screenshot 2019-06-11 at 2 25 30 PM

packages/teleport-plugin-react-styled-jsx/src/index.ts
```diff
@@ -59,13 +59,15 @@ export const createPlugin: ComponentPluginFactory<StyledJSXConfig> = (config) =>
...
-    const rootJSXNode = jsxNodesLookup[(node.content as UIDLElement).key] // TODO: Check for other types
-
     // We have the ability to insert the tag into the existig JSX structure, or do something else with it.
     // Here we take the JSX <style> tag and we insert it as the last child of the JSX structure
     // inside the React Component
-    rootJSXNode.children.push(jsxASTNodeReference)
+    const rootNodeKey = (node.content as UIDLElement).key
+    const rootJSXNode = rootNodeKey
+      ? jsxNodesLookup[rootNodeKey]
+      : jsxNodesLookup[Object.keys(jsxNodesLookup)[0]]
```
alexnm: this is a neat trick, but I wouldn't count on the fact that the object keys are in the right order. it also might be a bit harder for someone to understand. my proposal is to make a utility function that gives you the first element node in a given UIDL structure. might be useful in other cases in the future as well

JayaKrishnaNamburu: Sure, will change this 👍

packages/teleport-plugin-react-styled-jsx/src/index.ts
```diff
@@ -59,13 +59,19 @@ export const createPlugin: ComponentPluginFactory<StyledJSXConfig> = (config) =>
     }
 
     const jsxASTNodeReference = generateStyledJSXTag(styleJSXString.join('\n'))
-    const rootJSXNode = jsxNodesLookup[(node.content as UIDLElement).key] // TODO: Check for other types
-
     // We have the ability to insert the tag into the existig JSX structure, or do something else with it.
     // Here we take the JSX <style> tag and we insert it as the last child of the JSX structure
     // inside the React Component
-    rootJSXNode.children.push(jsxASTNodeReference)
+    let rootJSXNode
```
alexnm: It's easier to compute the rootJSXNode as the result of `traverseUIDLForRootElmNode` then you don't need to do the cast here nor the if statement

packages/teleport-plugin-react-styled-jsx/src/index.ts
```diff
@@ -59,13 +59,19 @@ export const createPlugin: ComponentPluginFactory<StyledJSXConfig> = (config) =>
...
+    const rootNodeKey = (node.content as UIDLElement).key
+    if (rootNodeKey) {
```
alexnm: only the else code is needed if actually, you don't need to check if the root is a UIDLElement as the function will check it

packages/teleport-shared/src/utils/uidl-utils.ts
```diff
@@ -424,3 +424,16 @@ export const transformAttributesAssignmentsToJson = (
 
   return newStyleObject
 }
+
+export const traverseUIDLForRootElmNode = (node) => {
```
alexnm: I would name this: `findFirstElementNode` because the traverse functions don't return something necessarily, they call a function on the node. in this case you are returning a specific node.

alexnm: also, add the typing for the `node`

packages/teleport-shared/src/utils/uidl-utils.ts
```diff
@@ -424,3 +424,16 @@ export const transformAttributesAssignmentsToJson = (
...
+  if (node && node.type) {
```
alexnm: why is this test needed?

packages/teleport-shared/src/utils/uidl-utils.ts
```diff
@@ -424,3 +424,16 @@ export const transformAttributesAssignmentsToJson = (
...
+    if (node.type === 'element') {
```
alexnm: this should be replaced with a switch like function:
 - case element -> return node
 - case static/dynamic throw error
 - case conditional/repeat/slot -> return recursive call

packages/teleport-shared/src/utils/uidl-utils.ts
JayaKrishnaNamburu: Not needed actually, I tried too many UIDL's and ended up doing this 😅 will remove 👍

packages/teleport-plugin-react-styled-jsx/src/index.ts
JayaKrishnaNamburu: Yeah right! I wonder how I missed this 😅

packages/teleport-shared/src/utils/uidl-utils.ts
JayaKrishnaNamburu: Yeah, I was doing this for the case of static and dynamic. I will throw an error from now on.

PR #272, opened 2019-06-12
Refactor/184 deprecate state identifier

### LINKED ISSUE #184: Deprecate StateIdentifier in favor of ConditionalIdentifier
The interface is used only for the React specific case of handling state management with hooks, would be better for consistency to have the new `ConditionalIdentifier` across all plugins. A `meta` object can be added for the `setter` value, since it is non-standard behavior (react state hook setter function)

ionutpasca:
closes #184

### DIFF
--- a/packages/teleport-plugin-react-base-component/src/utils.ts
+++ b/packages/teleport-plugin-react-base-component/src/utils.ts
@@ -320,9 +320,9 @@ const createReturnExpressionSyntax = (
 ) => {
   const returnStatement = t.returnStatement(jsxTagTree)
 
-  const stateHooks = Object.keys(stateDefinitions).map((stateKey) => {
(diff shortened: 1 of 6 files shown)


packages/teleport-plugin-react-base-component/src/utils.ts
```diff
@@ -393,31 +364,35 @@ const createStateChangeStatement = (
   }
 
   const stateKey = eventHandlerStatement.modifies
-  const stateIdentifier = stateIdentifiers[stateKey]
+  const stateDefinition = stateDefinitions[stateKey]
 
-  if (!stateIdentifier) {
+  if (!stateDefinition) {
```
alexnm: you can remove this if, as the validator is dealing with undeclared state keys

PR #276, opened 2019-06-13
fix(vue-base-com): Handling non-element root nodes for vue

### LINKED ISSUE #275: teleport-component-generator-vue malfunction?
teleport-component-generator-vue throws error: 
~~~
Expected node, not ` .... `
~~~

Reproducing scenario:
- use https://repl.teleporthq.io/
- enter the following UIDL sample:
~~~
{
  "name": "Static",
  "node": {
  	"type": "static",
	   "content":  ["Hello child"]
	 }
}
~~~
- check the difference between React and VUE component generators i.e. the React component generator provides component with string, the VUE compnent generator throws error 
~~~
generateReactComponent Error: Expected node, not `hello`
    at e.exports (index.js:1) .....
~~~
Note: the message **generateReactComponent** might be misguiding - it's the message inside the console.log.

Expected behavior:
- Both component generators should provide usable/code.

JayaKrishnaNamburu:
fixes #275


packages/teleport-plugin-vue-base-component/src/index.ts
```diff
@@ -28,13 +29,19 @@ export const createPlugin: ComponentPluginFactory<VueComponentConfig> = (config)
...
 
-    const templateContent = generateNodeSyntax(uidl.node, {
+    let templateContent = generateNodeSyntax(uidl.node, {
       templateLookup,
       dependencies,
       dataObject,
       methodsObject,
     })
 
+    if (typeof templateContent === 'string') {
```
JayaKrishnaNamburu: For every node we are adding a ``span`` as a wrapper expect in the case of static, where we are returning the string content itself. I tried adding in ``generateNodeSyntax`` itself, but there are edge cases where it will add two **span** tags. So added here, this will even work as a fall back when the template is returning a string, it will wrap in span so, in HAST generation, the pipeline won't break;

packages/teleport-component-generator-vue/__tests__/integration/component-with-conditional-root-array-child-node.json
```diff
@@ -0,0 +1,34 @@
+{
+    "$schema": "https://raw.githubusercontent.com/teleporthq/uidl-definitions/master/component.json",
+    "name": "ComponentWithConditionalArrayNode",
```
anamariaoros: I think we should stop adding json files inside the tests. There is an issue to completly eliminate them.
Instead, you can use the functions inside @teleporthq/teleport-shared/lib/builders/uidl-builders.
There are also examples of conditional components in `packages/teleport-component-generator-react/__tests__/integration/component-conditional.ts` or in `packages/teleport-component-generator-vue/__tests__/integration/component-conditional.ts`

packages/teleport-component-generator-vue/__tests__/integration/component-with-conditional-root-string-child-node.json
```diff
@@ -0,0 +1,27 @@
+{
+    "$schema": "https://docs.teleporthq.io/uidl-schema/v1/component.json",
+    "name": "ComponentWithConditionalRootNode",
```
anamariaoros: Same as in the previous .json file. This file can be removed and its content should generated with uidl-builders functions

packages/teleport-plugin-vue-base-component/src/index.ts
anamariaoros: Although the tests include a staticNode, the code is not covered. I think we should investigate why.

JayaKrishnaNamburu:
Screenshot 2019-07-01 at 11 31 43 AM

Hey, @anamariaoros I replaced the JSON fixtures, but the problem with the test coverage still persists. I think it is the behaviour of the tool, even if I remove the newly added lines. Test cases will start failing. And if we check the screenshot the **if** is coloured as ``green``, it means the line is being hitten at the time of execution.

@alexnm any ideas regarding this?

PR #278, opened 2019-06-14
feat(Attribute-mapping): Adding mapping for attributes

### LINKED ISSUE #188: Attribute mapping
There are cases in which the attribute keys might be mapped to something else, based on the flavor that is generated. Ex: `htmlFor`, `autoPlay`, etc.

There are two options:
- either we map these attributes inside the specific elements, but the downside is that we have to maintain that and we duplicate some mappings (ex: htmlFor)
- we create an additional layer inside the mapping, next to `elements` and `events` for `attributes`, where we map attribute names

anamariaoros:
closes #188

### DIFF
--- a/packages/teleport-component-generator/src/resolver/utils.ts
+++ b/packages/teleport-component-generator/src/resolver/utils.ts
@@ -68,13 +68,22 @@ export const resolveNavlinks = (uidlNode: UIDLNode, routesDefinition: UIDLStateD
 export const resolveNode = (uidlNode: UIDLNode, options: GeneratorOptions) => {
   traverseNodes(uidlNode, (node, parentNode) => {
     if (node.type === 'element') {
-      resolveElement(node.content, options)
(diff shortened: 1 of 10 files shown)


packages/teleport-component-generator/src/resolver/utils.ts
```diff
@@ -127,6 +132,20 @@ export const resolveElement = (element: UIDLElement, options: GeneratorOptions)
...
   }
 
+  if (originalElement.attrs && attributesMapping) {
+    const attrsKeys = Object.keys(originalElement.attrs)
+
+    attrsKeys
+      .filter((key) => attributesMapping[key])
+      .forEach((key) => {
+        originalElement.attrs = {
+          ...originalElement.attrs,
```
alexnm: you don't need to spread the originalElement.attrs here, only set the new value:
`originalElement.attrs[attributesMapping[key]] = originalElement.attrs[key]`

Issue #238, opened 2019-05-29
Default values for Array and Object should be returned as functions in Vue


JayaKrishnaNamburu:
Currently, the default values for Array and Object w.r.t to Vue needs a little bit of refactoring. Here is how we are currently generating

~~~
props: {
    items: {
      type: Array,
      default: [],
    },
  },
~~~
and we are assigning the values in the data object like this
~~~
data() {
    return {
      isVisible: true,
      items: ['angular', 'vue']
    };
  }
~~~
But Vue throws errors to use the default value itself, and is not picking up the values from data when we have an object that is colliding with the same name in data.

When we don't have any colliding variables we should declare the default values like this
~~~
props: {
  arr: {
    type: Array,
    default: () => []
  },
 user: {
   type: Object,
   default: () => ({})
 }
}
~~~

ionutpasca:
Hello @JayaKrishnaNamburu. This ticket will be closed by [this](https://github.com/teleporthq/teleport-code-generators/pull/279) PR.

The implementation from the PR solves the default value for Arrays and Objects returning them as functions and it also solves [this](https://github.com/teleporthq/teleport-code-generators/issues/210) issue.

The colliding related problem is fixed only for static data sources, used in the `repeat` type nodes, following to be fully fixed when we'll update the validation rules. @alexnm , please follow this.

PR #279, opened 2019-06-14
Fix/vue default values

ionutpasca:


ionutpasca:
closes #210 
closes #238

packages/teleport-component-generator/src/resolver/utils.ts
```diff
@@ -268,6 +270,22 @@ const isPowerOfTen = (value: number) => {
   return value === 1
 }
 
+export const ensureDataSourceUniqueness = (node: UIDLNode) => {
+  const staticRepeatsWithNoSource = countStaticRepeatsWithNoDataSource(node)
```
alexnm: you can remove this and use the index === 0 in the condition below

packages/teleport-plugin-vue-base-component/src/utils.ts
```diff
@@ -329,8 +333,16 @@ const createVuePropsDefinition = (uidlPropDefinitions: Record<string, UIDLPropDe
         )
     }
 
-    acc[name] =
-      typeof defaultValue !== 'undefined' ? { type: mappedType, default: defaultValue } : mappedType
+    let defaultPropValue = null
+
+    if (defaultValue !== 'undefined') {
+      const defaultValueAsFunc = new ParsedASTNode(
```
alexnm: as a small optimization thing, you could only compute defaultValueAsFunc when the type is `object` or `array`

PR #284, opened 2019-06-19
chore(Component generator): Deepmerge option for mergeMappings

### LINKED ISSUE #94: Deep merge for mappings
Currently when a mapping is added to the resolver, it is merged with object spread

```
export const mergeMappings = (oldMapping: Mapping, newMapping?: Mapping) => {
  if (!newMapping) {
    return oldMapping
  }

  return {
    elements: { ...oldMapping.elements, ...newMapping.elements },
    events: { ...oldMapping.events, ...newMapping.events },
  }
}
```

However, it may make sense to perform a deep merge for the mappings, as some of the attributes might be preserved.

This issue would require some analysis first, to see which are the use cases for a deep merge and which are the cases in which a merge with complete overwrite like in the current implementation makes sense. The deep merge could be an option to the `mergeMappings` function

andreiTnu:
closes #94

### DIFF
--- a/packages/teleport-component-generator/__tests__/resolver/utils.ts
+++ b/packages/teleport-component-generator/__tests__/resolver/utils.ts
@@ -18,7 +18,6 @@ import {
   UIDLStateDefinition,
   UIDLRepeatNode,
 } from '@teleporthq/teleport-types'
-import deepmerge = require('deepmerge')
 
 describe('generateUniqueKeys', () => {
   it('adds name and key to node', async () => {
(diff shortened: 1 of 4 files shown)


packages/teleport-component-generator/__tests__/resolver/utils.ts
```diff
@@ -10,13 +10,15 @@ import {
...
   ensureDataSourceUniqueness,
+  mergeMappings,
 } from '../../src/resolver/utils'
 import {
   UIDLElement,
   UIDLNode,
   UIDLStateDefinition,
   UIDLRepeatNode,
 } from '@teleporthq/teleport-types'
+import deepmerge = require('deepmerge')
```
alexnm: you don't need deepmerge in the tests here

PR #289, opened 2019-06-26
Fix/small updates in plugins

alexnm:


packages/teleport-component-generator/src/builder/index.ts
```diff
@@ -82,4 +82,10 @@ export default class Builder {
 
     return this.generators[type](content)
   }
+
+  // remove invalid chunks (which did not end up being created) from the linkAfter fields
+  // one use-case is when you want to remove the import plugin
+  private cleanupInvalidChunks(linkAfter: string[], chunks: ChunkDefinition[]) {
+    return linkAfter.filter((chunkName) => !!chunks.find((chunk) => chunk.name === chunkName))
```
ionutpasca: You could use here the `chunks.some(...` syntax, as you don't really need the returned value. This way you can also remove the `!!` syntax.

PR #290, opened 2019-06-26
fix(uidl): filename derived from sanitized component name

### LINKED ISSUE #235: Files names should be validated before assigning
If the UIDL name is having spaces in it, we are generated file names with spaces. We need to validate them before assigning.

Here is the reference --> https://github.com/teleporthq/teleport-code-generators/blob/master/packages/teleport-component-generator/src/index.ts#L68

This happens when the UIDL which we are using has its name like this

~~~
{
   "name": "Conditional Component"
}
~~~

alexnm:
the filename is now taken from the sanitized uidl name. also added a restriction on the elementType
to allow only alphanumeric characters, since the name is taken as a import variable when using
dependencies

fix #235

### DIFF
--- a/packages/teleport-uidl-validator/src/uidl-schemas/component.json
+++ b/packages/teleport-uidl-validator/src/uidl-schemas/component.json
@@ -266,7 +266,7 @@
       "additionalProperties": false,
       "properties": {
         "additionalProperties": false,
-        "elementType": {"type": "string"},
+        "elementType": {"type": "string", "pattern": "^([a-zA-Z0-9])*$"},
         "name": {"type": "string"},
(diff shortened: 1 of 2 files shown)


packages/teleport-uidl-validator/src/uidl-schemas/component.json
```diff
@@ -266,7 +266,7 @@
       "additionalProperties": false,
       "properties": {
         "additionalProperties": false,
-        "elementType": {"type": "string"},
+        "elementType": {"type": "string", "pattern": "^([a-zA-Z0-9])*$"},
```
JayaKrishnaNamburu: 👌

PR #298, opened 2019-07-02
repl-component: initial setup

### LINKED ISSUE #293: npm run lint error on repl-component sub-project
After I added repl-component sub-project, on ```npm run lint``` I recieved errors about dependencies from repl-component not added on package.json

As a workaround, I created a new branch *repl-component* and deactivated husky git hooks.

alexnm:
also closes #293


packages/teleport-repl-component/package.json
```diff
@@ -0,0 +1,44 @@
...
+  },
+  "devDependencies": {
+    "@stencil/core": "^1.0.7",
+    "rollup-plugin-node-builtins": "^2.1.2",
+    "rollup-plugin-node-globals": "^1.4.0"
+  },
+  "dependencies": {
+    "@teleporthq/teleport-component-generator-react": "^0.7.1",
+    "@teleporthq/teleport-component-generator-vue": "^0.7.0",
+    "codeflask": "^1.4.1"
```
JayaKrishnaNamburu: Hey, I think we should move ``@stencil/core`` and add ``prismjs`` to the **dependencies**

alexnm: prismjs yes, but stencil is a dev-dependency as it is not part of the bundle in the end

alexnm: actually, you are right, going to change it

JayaKrishnaNamburu: I see few ```rollup`` plugins which are being used in the ``stencil-config`` are also missing in package.json. Please cross check that too 👍

alexnm: it's a bit strange that these are required in `dependencies` and not `devDependencies`. after all they are used as `typescript` and `ts-node` in configuration file. the linter here is looking blindly at all the files, I will check if we can strictly analyze the `src` folders in all packages

alexnm:
I will merge this to have the code on master and to close #293. However, there's still a problem with using the latest version of the generators via lerna which I described here #303. Furthermore, I added some `repl` labelled issues.

PR #306, opened 2019-07-05
WIP: Feature/angular base component

JayaKrishnaNamburu:
Things to discuss/ todo.
- [ ] For ``componentGenerator`` if we should create the ``HTML`` as a new file or just map the generated HTML to ``templateUrl`` attribute.
- [ ] Mapping ``stateDefinition`` to ``Property Declaration`` (since the way state works and the way the two way binding for a value which is declared in component works the same)
- [ ] Mapping ``propDefinitions`` to ``Input Decorator``, (we have a catch where the input attribute mapping for the incoming props need to be adding in the HTML file)
- [ ] Support for *nested If* conditionals
- [ ] Support for *if-else* pattern in HTML file
- [ ] Event handling
- [ ] Support for slots, (In ``react`` we do `{props.children}`)
- [ ] Support for dynamic values in CSS using ``[ngClass] && [ngStyle]``
- [ ] Moving the ``fileName`` generation process into ``uidlResolve`` phase. (Unlike other flavors, in angular we need to map those names in the ``component.ts file``). So I feel we do the name generation in one place itself.
- [ ] Handling Styles

Generic Architectural changes to discuss

- [x] Moving the **AST to Code** generation phase to flavor specific component generator (So, we won't bloat ``teleport-shared`` with all types of builders)
- [x] Building two new packages by decoupling ``teleport-shared`` into ``teleport-typescript-builder`` and ``teleport-babel-builder``
- [ ] Check the feasibility and implementation of runtime caching dependencies as proposed by @vladnicula. So, we can solve the above two changes at once 🙌

### DIFF
--- a/packages/teleport-component-generator-angular/README.md
+++ b/packages/teleport-component-generator-angular/README.md
@@ -1,0 +1,11 @@
+# `@teleporthq/teleport-component-generator-angular`
+
+> TODO: description
+
+## Usage
+
+```
+const teleportComponentGeneratorAngular = require('@teleporthq/teleport-component-generator-angular');
+
+// TODO: DEMONSTRATE API
+```
--- a/packages/teleport-component-generator-angular/package.json
+++ b/packages/teleport-component-generator-angular/package.json
@@ -1,0 +1,34 @@
+{
+  "name": "@teleporthq/teleport-component-generator-angular",
+  "version": "0.8.1",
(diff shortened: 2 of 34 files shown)


JayaKrishnaNamburu:
Closing this as we are going to use `@babel/types` for this use case as well

PR #307, opened 2019-07-11
chore(watcher.js): A Watcher to that triggers for changes in file and rebuild the package

JayaKrishnaNamburu:
@alexnm this can help us in watching all packages at once and run ``tsc`` to build that package. I am using on my side to check if this creates any lag or something, since ``lerna run, build:watch --parallel`` is causing a lag for me.


package.json
```diff
@@ -24,7 +24,8 @@
     "version": "lerna version",
     "publish": "lerna publish from-package",
     "local:publish": "lerna exec -- npm publish --registry http://localhost:4873",
-    "local:unpublish": "lerna exec -- npm unpublish -f --registry http://localhost:4873"
+    "local:unpublish": "lerna exec -- npm unpublish -f --registry http://localhost:4873",
+    "watch": "node watcher.js"
```
alexnm: can we replace the `dev` task with this if it works better?

watcher.js
```diff
@@ -0,0 +1,25 @@
+const chokidar  = require('chokidar')
```
alexnm: can you please move this file into a `scripts` folder where we can further keep all setups and stuff like that?

JayaKrishnaNamburu: Sure 👍

package.json
JayaKrishnaNamburu: Yeah yeah, currently I want to test for a couple of days. So kept as a side script

scripts/watcher.js
```diff
@@ -0,0 +1,28 @@
...
+  log(chalk.redBright.bold(`${filePath} is changed`))
+
+  const path = filePath.split('/')
+  const location = `${path[0]}/${path[1]}/`
+
+  exec(`tsc`, { cwd: location}, (err, stdout) => {
+    if (!err ||  err === null) {
+      log(chalk.greenBright(`Package ${path[1]} is re-built`))
+    } else {
+      log(chalk.white(err))
```
JayaKrishnaNamburu: @alexnm I added these lines to throw errors when compilation fails, I tested this our for some time and its smooth enough.

alexnm: cool, eager to test this out also :)

JayaKrishnaNamburu: I am just pushing one more commit, by adding ``incremental`` in **tsconfig.json** that will decrease the compilation time. Please try with that

JayaKrishnaNamburu:
@alexnm you can check now

PR #309, opened 2019-07-15
fix(plugin-styles): Fix generation of empty styles

### LINKED ISSUE #288: Empty style block generates empty style code
The `React` style plugins all generate some sort of empty style object/component even though no keys are defined inside the `style` block.

Sample UIDL:
```json
{
  "name": "Simple Component",
  "propDefinitions": {
    "title": {
      "type": "string",
      "defaultValue": "Hello",
      "isRequired": true
    }
  },
  "node": {
    "type": "element",
    "content": {
      "elementType": "container",
      "style": {
        
      },
      "children": [
        {
          "type": "element",
          "content": {
            "elementType": "text",
            "children": [
              {
                "type": "dynamic",
                "content": {
                  "referenceType": "prop",
                  "id": "title"
                }
              },
              {
                "type": "static",
                "content": "World!"
              }
            ]
          }
        }
      ]
    }
  }
}
```

This can be easily tested in the repl: https://repl.teleporthq.io/

JayaKrishnaNamburu:
Fixes #288

### DIFF
--- a/packages/teleport-plugin-react-css-modules/src/index.ts
+++ b/packages/teleport-plugin-react-css-modules/src/index.ts
@@ -68,9 +68,20 @@ export const createPlugin: ComponentPluginFactory<ReactCSSModulesConfig> = (conf
       const { style, key } = element
       if (style) {
         const root = astNodesLookup[key]
-        const className = camelCaseToDashCase(key)
-        const classNameInJS = dashCaseToCamelCase(className)
         const { staticStyles, dynamicStyles } = splitDynamicAndStaticStyles(style)
+
+        if (Object.keys(staticStyles).length > 0) {
+          const className = camelCaseToDashCase(key)
+          const classNameInJS = dashCaseToCamelCase(className)
+
+          cssClasses.push(createCSSClass(className, getContentOfStyleObject(staticStyles)))
+
+          const classReferenceIdentifier = camelCaseClassNames
+            ? `styles.${classNameInJS}`
+            : `styles['${className}']`
+
+          addDynamicAttributeToJSXTag(root, 'className', classReferenceIdentifier)
+        }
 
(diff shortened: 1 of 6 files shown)


packages/teleport-plugin-react-css-modules/src/index.ts
```diff
@@ -66,7 +66,7 @@ export const createPlugin: ComponentPluginFactory<ReactCSSModulesConfig> = (conf
 
     traverseElements(uidl.node, (element) => {
       const { style, key } = element
-      if (style) {
+      if (style && Object.keys(style).length > 0) {
```
alexnm: can you check how I solved this in the stencil-css plugin which is still experimental? I think it's better to check both for inline styles as well as for the external styles:

https://github.com/teleporthq/teleport-code-generators/pull/308/files#diff-499a2b85fe4af0f3722b2a33b09ccad4R47

without this, if you have a dynamic style with css-modules, the class is still generated

packages/teleport-plugin-vue-css/src/index.ts
```diff
@@ -44,7 +44,7 @@ export const createPlugin: ComponentPluginFactory<VueStyleChunkConfig> = (config
     traverseElements(node, (element) => {
       const { style, key } = element
 
-      if (style) {
+      if (style && Object.keys(style).length > 0) {
```
alexnm: the same comment here as for css-modules. when splitting dynamic and static styles, we should check the separate objects so that we only generate what is actually needed

packages/teleport-plugin-react-css-modules/src/index.ts
JayaKrishnaNamburu: Sure

PR #311, opened 2019-07-16
(feat): Added a tab to show css file

### LINKED ISSUE #300: Show css or other output file when it is generated
The repl component can only show one code snippet, it would be great to have multiple tabs when for example `css-modules` are also generated.

JayaKrishnaNamburu:
Fixes #300

### DIFF
--- a/packages/teleport-repl-component/src/components/editor/editor.tsx
+++ b/packages/teleport-repl-component/src/components/editor/editor.tsx
@@ -25,21 +25,6 @@ export class Editor {
     this.setupEditor()
   }
 
-  public handleTabClick() {
-    const el = document.getElementById('css-tab')
-    this.showCssFile = !this.showCssFile
-    if (this.showCssFile) {
-      this.openTab()
-    } else {
-      el.classList.remove('show-css-tab')
-    }
-  }
-
-  public openTab() {
-    const el = document.getElementById('css-tab')
-    el.classList.add('show-css-tab')
-  }
-
   public render() {
     return (
       <div class="repl-wrapper" data-theme={this.dark ? 'dark' : ''}>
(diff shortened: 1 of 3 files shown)


packages/teleport-repl-component/src/components/editor/editor.tsx
```diff
@@ -16,13 +16,34 @@ export class Editor {
...
   }
 
+  public handleTabClick() {
+    const el = document.getElementById('css-tab')
+    const editor = document.getElementById('cssEditor')
+    this.showCssFile = !this.showCssFile
+    if (this.showCssFile) {
+      this.openTab()
+    } else {
+      editor.style.display = 'none'
```
alexnm: shouldn't these styles be set based on `showCssFile` inside the `render` function?

packages/teleport-repl-component/src/components/editor/editor.tsx
```diff
@@ -16,13 +16,34 @@ export class Editor {
...
+      el.style.height = '20px'
+    }
+  }
+
+  public openTab() {
+    const el = document.getElementById('css-tab')
+    const editor = document.getElementById('cssEditor')
+    editor.style.display = 'block'
```
alexnm: same question as above

packages/teleport-repl-component/src/components/editor/editor.tsx
JayaKrishnaNamburu: No, to get the render toggle functionality. On click of the header, I am listening to this func and setting the styles. I think stencil doesn't support for inline styles with state currently, so need to do using plain JS.

alexnm: the render function should be triggered any time you are changing something that is annotated with @State. so styles could be set based on that as well I think. have you tried it?

JayaKrishnaNamburu: Yes, inline styling like with state or prop is currently not supported, they support only class as of now

alexnm: then we could do class toggle, if state is X, add that class to the element

JayaKrishnaNamburu: Yeah makes sense, i will refactor by setting classes

PR #312, opened 2019-07-16
feat(build): experimental tasks for react generator

### LINKED ISSUE #310: Replace with build script instead of tsc
After merging the latest watcher, which waits for changes for files and builts it. We need to replace the **tsc** command with **build** command. Or, omit running build task for ``teleport-repl-component`` since repl have its own dev and build modes.

https://github.com/teleporthq/teleport-code-generators/blob/master/scripts/watcher.js#L20

alexnm:
closes #310


package.json
```diff
@@ -9,11 +9,10 @@
...
     "clean": "lerna run clean",
     "build": "lerna run build",
-    "dev": "node scripts/watcher.js",
-    "dev:core": "lerna run --parallel build:watch:core",
-    "dev:component": "lerna run --parallel build:watch:component",
-    "dev:project": "lerna run --parallel build:watch:project",
-    "dev:packer": "lerna run --parallel build:watch:packer",
+    "build:cjs": "lerna run build:cjs",
+    "build:esm": "lerna run build:esm",
+    "dev": "npm run build:cjs && npm run watch",
```
JayaKrishnaNamburu: Hey @alexnm can we change these two lines with this
~~~
     "build": "lerna run build --parallel",
    "build:cjs": "lerna run build:cjs --parallel",
    "build:esm": "lerna run build:esm --parallel",
~~~

JayaKrishnaNamburu: This is decreasing the build time

alexnm: sure, can you do a PR on this?

JayaKrishnaNamburu: Ha sure, I will update

PR #315, opened 2019-07-24
refactor(package.json): Running build script in parallel mode

JayaKrishnaNamburu:


package.json
```diff
@@ -8,9 +8,9 @@
...
     "lint": "tslint -c tslint.json 'packages/teleport-*/src/**/*.ts{,x}'",
     "clean": "lerna run clean",
-    "build": "lerna run build",
-    "build:cjs": "lerna run build:cjs",
-    "build:esm": "lerna run build:esm",
+    "build": "lerna run build --parallel",
+    "build:cjs": "lerna run build:cjs --parallel",
+    "build:esm": "lerna run build:esm --parallel",
     "dev": "npm run build:cjs && npm run watch",
     "watch": "node scripts/watcher.js",
```
JayaKrishnaNamburu: Any specific reason for using ``npm run build`` instead of ``yarn build`` ?

package.json
```diff
@@ -7,7 +7,7 @@
   "description": "Code generators and UIDL definition library",
   "scripts": {
     "lint": "tslint -c tslint.json 'packages/teleport-*/src/**/*.ts{,x}'",
-    "clean": "lerna run clean",
+    "clean": "lerna run clean --parallel",
     "build": "lerna run build",
```
JayaKrishnaNamburu: I removed --parallel from build scripts and here are the reasons

1. Our packages are dependent on one another, so when you run ``yarn clean`` and run build in parallel mode, it breaks. Because the build script is running in all packages at once and we cannot resolve the import step because all the dist folders are removed. Happens only in specific packages where we explicitly import saying ``<package-name>/cjs/index``

2. We currently reached 37 packages, running the build in all these at once is pushing CPU usage and causing lag in the machine.

PR #318, opened 2019-07-25
fix(styled-comp): Converting to camel case when passing as props

### LINKED ISSUE #316: Dash to camel case needed for styled-components plugin
While testing out some stuff, I noticed this bug with the styled-components plugin, when using dynamic values in styles and passing them down to the `styled` component.

Screenshot 2019-07-24 at 14 43 01

JayaKrishnaNamburu:
fixes #316


packages/teleport-uidl-validator/src/uidl-schemas/component.json
```diff
@@ -45,7 +45,7 @@
     "propDefinitions": {
       "type": "object",
       "patternProperties": {
-        ".*": {
+        "^([a-zA-Z0-9])*$": {
```
JayaKrishnaNamburu: I tried passing special characters too, but our validator is not throwing errors when we pass special characters in UIDL

alexnm: I think patternProperties is acting a bit differently here maybe? what strings have you tried?

alexnm: also, do we need this for the prop definition? since the props we're worried about are usually added from the style rules when using styled-components

JayaKrishnaNamburu: No, the regex is correct, I tested online and in another schema's it is working fine. we need this to eliminate special-characters being used in the prop name. The issue we encountered using ``background-color`` as a prop name

alexnm: but background-color was not defined in propDefinitions, it is just passed to the styled component to be able to use dynamic values

JayaKrishnaNamburu:
@alexnm reverted the change in the schema.

PR #330, opened 2019-08-01
refactor(vue): extract html generation syntax into shared package

alexnm:


alexnm:
@JayaKrishnaNamburu this is touching a lot of files but it's mostly renaming + the moving of the generate html syntax functions into the shared package

alexnm:
> Looks Good 
> 
> ```
> createHTMLTemplateSyntax(uidl.node, {
>       templateLookup,
>       dependencies,
>       dataObject,
>       methodsObject,
> 	  flavour    
> })
> ```
> 
> I think `createHTMLTemplateSyntax` should accept another attribute like this and so we can use the template specific tags. I will handle this in `angular-base`

I would create a 3rd parameter, HtmlGenerstion

> Looks Good 
> 
> ```
> createHTMLTemplateSyntax(uidl.node, {
>       templateLookup,
>       dependencies,
>       dataObject,
>       methodsObject,
> 	  flavour    
> })
> ```
> 
> I think `createHTMLTemplateSyntax` should accept another attribute like this and so we can use the template specific tags. I will handle this in `angular-base`

It's better to pass a new object as the 3rd argument, HtmlTemplateGenerationOptions, which has the flavor and maybe other options/flags.

PS: let's use the american "flavor" instead of the british one :)

alexnm:
maybe we can even pass the template rules as params.

like for dynamic - vue is `:${key}`, for angular it's `[${key}]`
for events - vue is @ in front, angular is () and so on

and you can specify the same for conditionals and loops and slots, etc

JayaKrishnaNamburu:
> maybe we can even pass the template rules as params.
> 
> like for dynamic - vue is `:${key}`, for angular it's `[${key}]`
> for events - vue is @ in front, angular is () and so on
> 
> and you can specify the same for conditionals and loops and slots, etc

Yeah i feel we have have a object map something like this

{
angular: { event:""}
react: { event: "" }
}

PR #331, opened 2019-08-02
refactor(core): template generation syntax is now configurable from base plugins

alexnm:


packages/teleport-shared/src/node-handlers/node-to-html/constants.ts
```diff
@@ -0,0 +1,15 @@
...
+  eventBinding: (value) => `@${value}`,
+  valueBinding: (value) => `:${value}`,
+  eventEmmitter: (value) => `this.$emit('${value}')`,
+  conditionalAttr: 'v-if',
+  repeatAttr: 'v-for',
+  repeatIterator: (iteratorName, iteratedCollection, useIndex) => {
+    const iterator = useIndex ? `(${iteratorName}, index)` : iteratorName
+    return `${iterator} in ${iteratedCollection}`
+  },
+}
```
JayaKrishnaNamburu: Are we considering this is as a default fallback? If we are not sending the mappings from `createHTMLTemplateSyntax` from the base component?

packages/teleport-plugin-vue-base-component/src/index.ts
```diff
@@ -31,12 +31,27 @@ export const createPlugin: ComponentPluginFactory<VueComponentConfig> = (config)
...
+        interpolation: (value) => `{{ ${value} }}`,
+        eventBinding: (value) => `@${value}`,
+        valueBinding: (value) => `:${value}`,
+        eventEmmitter: (value) => `this.$emit('${value}')`,
+        conditionalAttr: 'v-if',
+        repeatAttr: 'v-for',
+        repeatIterator: (iteratorName, iteratedCollection, useIndex) => {
+          const iterator = useIndex ? `(${iteratorName}, index)` : iteratorName
+          return `${iterator} in ${iteratedCollection}`
+        },
```
JayaKrishnaNamburu: How about moving to a separate file like `DEFAULT_TEMPLATE_SYNTAX` to make this look short and crisp

alexnm: I wanted to have this visible in the main plugin, it's not a constant necessarily, as it has functions as well, it's the configuration of the template syntax function

PR #338, opened 2019-08-09
refactor(css): teleport-plugin-css now works for both html templates and jsx code chunks

### LINKED ISSUE #332: Plugin for React external stylesheets
There are still considerable numbers of React users that do not use css-modules or css-in-js.

For that, we could have a plugin that extracts the css in a separate file next to the .js file and imports the style file: `import './component.css'`

alexnm:
`react` now supports: `CSS`, `CSSModules`, `InlineStyles`, `StyledJSX`, `StyledComponents`, `JSS`
`preact` now supports: `CSS`, `CSSModules`, `InlineStyles`

also closes #332


packages/teleport-plugin-angular-base-component/src/utils.ts
```diff
@@ -16,11 +13,7 @@ export const generateComponentDecorator = (uidl: ComponentUIDL, t = types) => {
       t.objectProperty(t.identifier('selector'), t.stringLiteral('app-root')),
       t.objectProperty(
         t.identifier('templateUrl'),
-        t.stringLiteral(`./${getComponentFileName(uidl)}.${FILE_TYPE.HTML}`)
-      ),
-      t.objectProperty(
```
JayaKrishnaNamburu: Are we going to handle path in project generators alex ?

alexnm: no, I moved this to the css-plugin

JayaKrishnaNamburu: But the output file is not referring I think

```ts
import { Component } from '@angular/core'

@Component({
  selector: 'app-root',
  templateUrl: 'contact-form.html',
  styleUrls: ['contact-form.css'],
})
export class ContactForm {
  constructor() {}
}
```

JayaKrishnaNamburu: And if the css is moved, the template should still follow the same for `tempalteUrl` right --> ``templateUrl: './contact-form.html'``

PR #341, opened 2019-08-13
Feature/stencil project generator

JayaKrishnaNamburu:
*Stencil Generator* is almost completed, but here are the two limitations that we need to look into

- There are few reserve keywords in Stencil like `title`. we need to make a list and cross-check when the UIDL contains such declarations
- We add attributes directly to the `jsx` tags like 
```
<textaread rows="8" cols="50" />
```
Since we have type-checking enabled in the stencil, binding numeric characters like above is not supported. Instead, it should be

```
<textarea rows={8} cols={50}
```

### DIFF
--- a/packages/teleport-plugin-stencil-app-routing/README.md
+++ b/packages/teleport-plugin-stencil-app-routing/README.md
@@ -1,0 +1,11 @@
+# `@teleporthq/teleport-plugin-stencil-app-routing`
+
+> TODO: description
+
+## Usage
+
+```
+const teleportPluginStencilAppRouting = require('@teleporthq/teleport-plugin-stencil-app-routing');
+
+// TODO: DEMONSTRATE API
+```
--- a/packages/teleport-plugin-stencil-app-routing/README.md
+++ b/packages/teleport-plugin-stencil-app-routing/README.md
@@ -1,11 +1,14 @@
 # `@teleporthq/teleport-plugin-stencil-app-routing`
 
-> TODO: description
+A plugin for generating the routing file for a Stencil project.
 
-## Usage
+> This package is part of the [teleport ecosystem](https://github.com/teleporthq/teleport-code-generators). For a complete guide, check out the [official documentation](https://docs.teleporthq.io/).
 
+## Install
+```bash
+npm install @teleporthq/teleport-plugin-stencil-app-routing
 ```
-const teleportPluginStencilAppRouting = require('@teleporthq/teleport-plugin-stencil-app-routing');
-
-// TODO: DEMONSTRATE API
+or
+```bash
+yarn add @teleporthq/teleport-plugin-stencil-app-routing
 ```
--- a/packages/teleport-plugin-stencil-app-routing/src/index.ts
+++ b/packages/teleport-plugin-stencil-app-routing/src/index.ts
@@ -44,13 +44,13 @@ export const createPlugin: ComponentPluginFactory<StencilRouterConfig> = (config
       linkAfter: [importChunkName],
     })
 
-    const exportClassAST = createClassDecleration(routes, routeDefinitions)
+    const classDeclarationAST = createClassDecleration(routes, routeDefinitions)
 
     chunks.push({
       name: componentChunkName,
       type: CHUNK_TYPE.AST,
       fileType: FILE_TYPE.TSX,
-      content: exportClassAST,
+      content: classDeclarationAST,
       linkAfter: [componentDecoratorChunkName],
     })
 
--- a/packages/teleport-plugin-stencil-app-routing/src/index.ts
+++ b/packages/teleport-plugin-stencil-app-routing/src/index.ts
@@ -34,7 +34,11 @@ export const createPlugin: ComponentPluginFactory<StencilRouterConfig> = (config
 
     /* The name should be injected only with AppRoot only then it acts as entry point,
     Sending only Root because app is appended while generation of decorators*/
-    const decoratorAST = createComponentDecorator('Root')
+    const params = {
+      tag: 'app-root',
+      shadow: true,
+    }
+    const decoratorAST = createComponentDecorator(params)
 
     chunks.push({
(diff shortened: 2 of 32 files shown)


packages/teleport-plugin-stencil-app-routing/README.md
```diff
@@ -0,0 +1,11 @@
+# `@teleporthq/teleport-plugin-stencil-app-routing`
```
alexnm: don't forget to complete this

packages/teleport-plugin-stencil-app-routing/src/index.ts
```diff
@@ -0,0 +1,63 @@
...
+
+    chunks.push({
+      name: componentDecoratorChunkName,
+      type: CHUNK_TYPE.AST,
+      fileType: FILE_TYPE.TSX,
+      content: [decoratorAST],
+      linkAfter: [importChunkName],
+    })
+
+    const exportClassAST = createClassDecleration(routes, routeDefinitions)
```
alexnm: classDeclaration

packages/teleport-project-generator-stencil/README.md
```diff
@@ -0,0 +1,11 @@
+# `teleport-project-generator-stencil`
```
alexnm: same reminder to complete the info

packages/teleport-project-generator-stencil/src/index.ts
```diff
@@ -0,0 +1,70 @@
...
+      path: ['src', 'components'],
+      fileName: 'app-root',
+    },
+    entry: {
+      generator: htmlFileGenerator,
+      appRootOverride: `<app-root></app-root>`,
+      path: ['src'],
+      fileName: 'index',
+      customScriptTags: [
+        { type: 'module', path: ['build', 'app.esm.js'] },
```
alexnm: I would keep path: string[] only when we refer to folders, not also files. I see two options:
1. path is ['build'] and there's another field fileName: 'app.esm.js'
2. path is a simple string: (build/app.esm.js). maybe this one makes more sense since the path is always the same when you know the build strategy of the project

packages/teleport-project-generator/src/file-handlers.ts
```diff
@@ -182,6 +191,31 @@ const createHTMLEntryFileChunks = (uidl: ProjectUIDL, options: EntryFileOptions)
     }
   })
 
+  // Stencil need to inject and point out the generated build files
```
alexnm: this functionality can be used by any project generator, I think it's ok to have it as generic as it is, without explaining the specific need for stencil, so you can change into: `For frameworks that need to ....`

packages/teleport-shared/src/node-handlers/node-to-jsx/index.ts
```diff
@@ -88,7 +88,11 @@ const generateElementNode = (
 ) => {
   const { dependencies, nodesLookup } = params
   const { elementType, children, key, attrs, dependency, events } = node.content
-  const elementTag = createJSXTag(elementType)
+  // Append app for only local dependencies
+  const elementTag =
```
alexnm: check lines 122-132 where we handle dependencies. based on the `dependencyHandling` mode we do some things that can overlap with this functionality. let's talk about it to have a single solution maybe for handling the names of the local dependencies

packages/teleport-shared/src/utils/ast-jsx-utils.ts
```diff
@@ -146,3 +147,16 @@ export const renameJSXTag = (jsxTag: types.JSXElement, newName: string, t = type
     jsxTag.closingElement.name = t.jsxIdentifier(newName)
   }
 }
+
+export const createComponentDecorator = (name: string, t = types) => {
```
alexnm: if we pass the `{ tag: ..., shadow: true}` object as a param, can't we reuse this for angular and stencil?

JayaKrishnaNamburu: Yeah, we can sent the props and the function can take care of looping and generating

packages/teleport-project-generator-stencil/src/index.ts
JayaKrishnaNamburu: Makes sense, made changes 👍

packages/teleport-project-generator-stencil/src/index.ts
```diff
@@ -0,0 +1,70 @@
...
+      },
+    },
+    router: {
+      generator: routingComponentGenerator,
+      path: ['src', 'components'],
+      fileName: 'app-root',
+    },
+    entry: {
+      generator: htmlFileGenerator,
+      appRootOverride: `<app-root></app-root>`,
```
alexnm: maybe we should move appRootOverride, customScriptTags and customLinkTags into an `option object`?

packages/teleport-shared/src/node-handlers/node-to-jsx/index.ts
```diff
@@ -88,7 +88,13 @@ const generateElementNode = (
 ) => {
   const { dependencies, nodesLookup } = params
   const { elementType, children, key, attrs, dependency, events } = node.content
-  const elementTag = children ? createJSXTag(elementType) : createSelfClosingJSXTag(elementType)
+  // Append app for only local dependencies
+  const elementTag =
+    dependency && dependency.type === 'local' && options.customElementTag
+      ? createJSXTag(options.customElementTag(elementType))
```
alexnm: maybe here you can compute the name of the tag before and then leave the same logic as above (previous line 91) where the tag is created either as a selfclosing or regular JSX tag

packages/teleport-project-generator-stencil/src/index.ts
JayaKrishnaNamburu: Yes, that way it stays in sync with all the others

JayaKrishnaNamburu:
@alexnm you can merge this now 👍 🚀

PR #348, opened 2019-08-22
fix(preact-generator): Picking a inbuilt index template and injecting custom tags

### LINKED ISSUE #342: Preact project does not use the generated index.html
While testing out the preact project, I noticed that the roboto font and the title are not set in the running project. then I looked at the build task in preact and it is creating a new file, not using the existing one. In React this works, values set in index.html are preserved when running the app

JayaKrishnaNamburu:
…jecting custom tags

closes #342

### DIFF
--- a/packages/teleport-project-generator-preact/src/constants.ts
+++ b/packages/teleport-project-generator-preact/src/constants.ts
@@ -1,0 +1,17 @@
+export const customHeadContent = () => `<% if (htmlWebpackPlugin.options.manifest.theme_color) { %>
+	<meta name="theme-color" content="<%= htmlWebpackPlugin.options.manifest.theme_color %>">
+<% } %>
+<% for (var chunk of webpack.chunks) { %>
+ <% if (chunk.names.length === 1 && chunk.names[0] === 'polyfills') continue; %>
+	<% for (var file of chunk.files) { %>
+		<% if (htmlWebpackPlugin.options.preload && file.match(/\.(js|css)$/)) { %>
+			<link rel="preload" href="<%= htmlWebpackPlugin.files.publicPath + file %>" as="<%= file.match(/\.css$/)?'style':'script' %>">
+		<% } else if (file.match(/manifest\.json$/)) { %>
+			<link rel="manifest" href="<%= htmlWebpackPlugin.files.publicPath + file %>">
+		<% } %>
+	<% } %>
+<% } %>`
+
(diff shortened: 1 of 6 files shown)


packages/teleport-project-generator-preact/src/constants.ts
```diff
@@ -0,0 +1,17 @@
+export const customHeadContent = () => `<% if (htmlWebpackPlugin.options.manifest.theme_color) { %>
```
alexnm: you can keep these as strings if they don't have any dynamic parts

packages/teleport-project-generator-preact/src/index.ts
```diff
@@ -49,6 +47,21 @@ export const createPreactProjectGenerator = () => {
...
       path: ['src'],
       fileName: 'index',
+      options: {
+        appRootOverride: customBodyContent(),
+        customHeadContent: customHeadContent(),
+        customScriptTags: [
+          {
+            target: 'body',
+            type: 'defer',
+            path: `<%= htmlWebpackPlugin.files.chunks['bundle'].entry %>`,
```
alexnm: isn't this the root actually? and this:
```
<%= htmlWebpackPlugin.options.ssr({
	url: '/'
}) %>
```

would be another custom script tag?

packages/teleport-project-generator/src/file-handlers.ts
```diff
@@ -119,18 +127,31 @@ const createHTMLEntryFileChunks = (uidl: ProjectUIDL, options: EntryFileOptions)
...
+
+      const targetNode = target === 'body' ? bodyNode : headNode
       const scriptTag = createHTMLNode('script')
-      if (type === 'module') {
-        addAttributeToNode(scriptTag, 'type', type)
-      } else {
-        addAttributeToNode(scriptTag, 'nomodule', '')
+
+      if (type) {
+        type === 'module'
```
alexnm: can you please explain the logic here in a comment?

JayaKrishnaNamburu: This needs a small refactor, the current logic won't work for all edge cases. I will update this now 👍

packages/teleport-project-generator-preact/src/index.ts
JayaKrishnaNamburu: Yes, the SSR tag is outside the script and this is added inside a script tags

PR #351, opened 2019-08-26
fix(angular-base): Handling multiple statements in event handlers

### LINKED ISSUE #350: Handling multiple statements in event handlers
When adding more than one statement in an event handler:
Screenshot 2019-08-22 at 16 40 12

The template part is properly updated to call a function, but the function is not implemented/added on the component class

Screenshot 2019-08-22 at 16 40 04

Also notice the propCall is not handled, it uses the same syntax as vue, I think this is the solution we're looking for: https://angular.io/api/core/EventEmitter

JayaKrishnaNamburu:
Fixes #350

### DIFF
--- a/packages/teleport-plugin-angular-base-component/src/utils.ts
+++ b/packages/teleport-plugin-angular-base-component/src/utils.ts
@@ -2,16 +2,31 @@ import * as types from '@babel/types'
 import {
   convertValueToLiteral,
   getTSAnnotationForType,
+  createMethodsObject,
 } from '@teleporthq/teleport-shared/dist/cjs/utils/ast-js-utils'
-import { UIDLPropDefinition, UIDLStateDefinition } from '@teleporthq/teleport-types'
+import {
+  UIDLPropDefinition,
+  UIDLStateDefinition,
+  UIDLEventHandlerStatement,
+} from '@teleporthq/teleport-types'
 
 export const generateExportAST = (
   componentName: string,
   propDefinitions: Record<string, UIDLPropDefinition>,
   stateDefinitions: Record<string, UIDLStateDefinition>,
   dataObject: Record<string, any>,
+  methodsObject: Record<string, UIDLEventHandlerStatement[]>,
   t = types
 ) => {
+  let angularMethodsAST = []
+  if (Object.keys(methodsObject).length > 0) {
+    angularMethodsAST = createMethodsObject(
+      methodsObject,
+      propDefinitions,
+      'angular'
+    ) as types.ClassMethod[]
+  }
+
   const propDeclaration = Object.keys(propDefinitions).map((propKey) =>
     t.classProperty(
       t.identifier(propKey),
@@ -43,6 +58,7 @@ export const generateExportAST = (
       ...propertyDecleration,
       ...dataDeclaration,
       constructorAST(),
+      ...angularMethodsAST,
     ])
   }
 
--- a/packages/teleport-plugin-angular-base-component/src/utils.ts
+++ b/packages/teleport-plugin-angular-base-component/src/utils.ts
@@ -27,14 +27,28 @@ export const generateExportAST = (
     ) as types.ClassMethod[]
   }
 
-  const propDeclaration = Object.keys(propDefinitions).map((propKey) =>
-    t.classProperty(
+  const propDeclaration = Object.keys(propDefinitions).map((propKey) => {
+    const definition = propDefinitions[propKey]
+    if (definition.type === 'func') {
+      return t.classProperty(
+        t.identifier(propKey),
+        t.newExpression(t.identifier('EventEmitter'), []),
+        t.typeAnnotation(
+          t.genericTypeAnnotation(
+            t.identifier('EventEmitter'),
+            t.typeParameterInstantiation([t.anyTypeAnnotation()])
+          )
+        ),
+        [t.decorator(t.callExpression(t.identifier('Output'), []))]
+      )
+    }
+    return t.classProperty(
       t.identifier(propKey),
       convertValueToLiteral(propDefinitions[propKey].defaultValue),
       t.tsTypeAnnotation(getTSAnnotationForType(propDefinitions[propKey].type)),
       [t.decorator(t.callExpression(t.identifier('Input'), []))]
     )
-  )
+  })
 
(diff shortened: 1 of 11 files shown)


packages/teleport-plugin-angular-base-component/src/utils.ts
```diff
@@ -2,24 +2,53 @@ import * as types from '@babel/types'
...
+      methodsObject,
+      propDefinitions,
+      'angular'
+    ) as types.ClassMethod[]
+  }
+
+  const propDeclaration = Object.keys(propDefinitions).map((propKey) => {
+    const definition = propDefinitions[propKey]
+    if (definition.type === 'func') {
+      return t.classProperty(
```
JayaKrishnaNamburu: @alexnm  what is your opinion. I am not sure if it is a safe assumption to consider all props which are functions are event-based ones.

Repl Link --> https://repl.teleporthq.io/?uidlLink=qoaic5fuhhl6w8hgej185h
Code sandbox link --> https://codesandbox.io/s/angular-6zw59

packages/teleport-shared/src/utils/ast-js-utils.ts
```diff
@@ -111,3 +112,85 @@ export const getTSAnnotationForType = (type: any, t = types) => {
       return t.tsUnknownKeyword()
   }
 }
+
+export const createMethodsObject = (
+  methods: Record<string, UIDLEventHandlerStatement[]>,
+  propDefinitions: Record<string, UIDLPropDefinition>,
+  flavor: 'angular' | 'vue',
```
alexnm: this function is not truly shareable, since it has the flavor dependency. I would handle everything related to the specifics of vue and angular in their base plugins and leave here in shared these functions:

createEventHandlerBody
createStateChangeStatement
createPropCallStatement

packages/teleport-shared/src/utils/ast-js-utils.ts
```diff
@@ -111,3 +112,85 @@ export const getTSAnnotationForType = (type: any, t = types) => {
...
+
+  if (!propDefinition) {
+    console.warn(`No prop definition was found for function "${propFunctionKey}"`)
+    return null
+  }
+
+  // In vue it's favorable to use $emit for a specific event than sending the function as a prop
+  const emitter =
+    flavor === 'angular'
+      ? t.expressionStatement(
```
alexnm: can this logic be sent as a parameter of how to handle the prop call / event emitter thing?

packages/teleport-shared/src/node-handlers/node-to-html/constants.ts
```diff
@@ -6,6 +6,7 @@ export const DEFAULT_TEMPLATE_SYNTAX: HTMLTemplateSyntax = {
   eventBinding: (value) => `@${value}`,
   valueBinding: (value) => `:${value}`,
   eventEmmitter: (value) => `this.$emit('${value}')`,
+  eventNameBinding: (value) => `${value}`,
```
alexnm: this is equivalent to (value) => value

packages/teleport-project-packer-test/src/index.ts
```diff
@@ -22,7 +22,7 @@ import {
   STENCIL_GITHUB_PROJECT,
 } from './constants'
 
-import projectUIDL from '../../../examples/uidl-samples/project.json'
+import projectUIDL from '../../../examples/test-samples/project-sample.json'
```
alexnm: make sure you revert this when we finish

packages/teleport-plugin-angular-base-component/src/utils.ts
alexnm: I think this works for now, please write a comment about this as well, so we know in the future

packages/teleport-plugin-angular-base-component/src/utils.ts
```diff
@@ -2,24 +2,53 @@ import * as types from '@babel/types'
...
 ) => {
-  const propDeclaration = Object.keys(propDefinitions).map((propKey) =>
-    t.classProperty(
+  let angularMethodsAST = []
+  if (Object.keys(methodsObject).length > 0) {
+    angularMethodsAST = createMethodsObject(
+      methodsObject,
+      propDefinitions,
+      'angular'
+    ) as types.ClassMethod[]
```
alexnm: type cast could be removed once the function is refactored to return only the body of each method

packages/teleport-shared/src/node-handlers/node-to-html/types.ts
```diff
@@ -15,6 +15,7 @@ export interface HTMLTemplateSyntax {
   interpolation?: (value: string) => string
   valueBinding?: (value: string, node?: UIDLElementNode) => string
   eventBinding?: (value: string) => string
+  eventNameBinding?: (value: string) => string
```
alexnm: I'd find a different name for this, this is just used to add the function call parentheses right? how about: `eventHandlersBindingMode`: 'reference' | 'call' and maybe leave some explanations

packages/teleport-shared/src/utils/ast-js-utils.ts
JayaKrishnaNamburu: Yeah, felt the same but wanted to test the solution first if it really works. Now I will clean this up 👍 :D

packages/teleport-shared/src/utils/ast-js-utils.ts
JayaKrishnaNamburu: I moved the prop calls to two separate individual functions in angular-base and vue-base. It looks good that way as we don't need to pass flavor flags anymore.

packages/teleport-shared/src/node-handlers/node-to-html/types.ts
JayaKrishnaNamburu: Regarding this, we can't use reference, callee pattern. Here is the reason, if we combine and make it as a single `templateSyntax.eventBindingMode` which return 
```
@click="handleClick"
```
or
```
(click)="handleClick()"
```
according to syntax which looks clean but the *prettyHTML* which we use will replace special characters and spaces.
so it ends up generating 
```
(click)<#&..some constants>"handleClick"
```

Issue #355, opened 2019-08-29
Getting the playground-packer ready for integration


alexnm:
Configuring the playground-packer with all the generators and capabilities so that it can be fully integrated in the playground

- [x] setup all project generators
- [x] decide on how to handle the templates for the projects
- [x] export all enums and strings so that they can be used in the playground
- [x] export mappings if they need to be used outside the core generators
- [ ] ~~export the UIDL validation function for project and component generators~~

Still debatable
- [x] packer playground should export the component generators to allow code preview to work
- [ ] the repl web component can be integrated into the playground

PR #357, opened 2019-08-29
feat(angular-module-gen): Project Generator for Angular using default angular-cli template

JayaKrishnaNamburu:


packages/teleport-plugin-module-generator-angular/package.json
```diff
@@ -0,0 +1,32 @@
+{
+  "name": "@teleporthq/teleport-plugin-module-generator-angular",
```
alexnm: Based on our naming conventions, this should be: `teleport-plugin-angular-module-generator`

packages/teleport-plugin-module-generator-angular/src/index.ts
```diff
@@ -0,0 +1,127 @@
...
+import {
+  createRoutesAST,
+  createPageRouteAST,
+  createExportModuleAST,
+  constructLocalDependency,
+  createRootModuleDecorator,
+  constructComponentDependency,
+  createComponentModuleDecorator,
+  createPageModuleModuleDecorator,
+  constructRouteForComponentsModule,
```
alexnm: nice arrangement, @anamariaoros would be proud 🤣

packages/teleport-plugin-module-generator-angular/src/index.ts
```diff
@@ -0,0 +1,127 @@
...
+          dependencies.ComponentsModule = constructRouteForComponentsModule('../..')
+          dependencies.CommonModule = ANGULAR_COMMON_MODULE
+          const componentName = `${uidl.name}Component`
+          dependencies[componentName] = constructLocalDependency(uidl.meta.fileName)
+
+          routesAST = createPageRouteAST(componentName)
+          ngModuleAST = createPageModuleModuleDecorator(componentName)
+          moduleDecoratorAST = createExportModuleAST(uidl.meta.moduleName)
+
+          // Acording to widely followd convention module should have .module in its name
```
alexnm: followed

packages/teleport-plugin-module-generator-angular/src/index.ts
```diff
@@ -0,0 +1,127 @@
...
+          uidl.meta.fileName = `${uidl.meta.fileName}.module`
+        }
+        break
+      case 'component':
+        {
+          dependencies.CommonModule = ANGULAR_COMMON_MODULE
+          // Looping throgh all components and adding importing them into component module
```
alexnm: "through" and remove "adding" from the sentence

packages/teleport-plugin-module-generator-angular/src/utils.ts
```diff
@@ -0,0 +1,172 @@
...
+        ]),
+      ])
+    ),
+  ])
+}
+
+export const constructLocalDependency = (componentName: string) => {
+  const dependency: UIDLDependency = {
+    type: 'local',
+    path: `./${componentName}.component`,
```
alexnm: are these dependencies only inside the component folder? so they are not affected by the UIDL folder specification and the project strategy?

packages/teleport-project-generator-angular/src/index.ts
```diff
@@ -0,0 +1,98 @@
...
+import { createComponentGenerator } from '@teleporthq/teleport-component-generator'
+import { createProjectGenerator } from '@teleporthq/teleport-project-generator'
+import { FILE_TYPE } from '@teleporthq/teleport-shared/dist/cjs/constants'
+import { Mapping } from '@teleporthq/teleport-types'
+
+import prettierHTML from '@teleporthq/teleport-postprocessor-prettier-html'
+import { createPlugin as createImportPlugin } from '@teleporthq/teleport-plugin-import-statements'
+
+import { CUSTOM_BODY_CONTENT } from './constants'
+import anuglarProjectMapping from './angular-mapping.json'
```
alexnm: angular

packages/teleport-project-generator-angular/src/index.ts
```diff
@@ -0,0 +1,98 @@
...
+  const rootModuleGeneratorAngular = createAngularModuleGenerator({ moduleType: 'root' })
+  const componentModuleGeneratorAngular = createAngularModuleGenerator({ moduleType: 'component' })
+  const pagesModuleGeneratorAngular = createAngularModuleGenerator({ moduleType: 'page' })
+
+  const angularComponentGenerator = createAngularComponentGenerator()
+  angularComponentGenerator.addMapping(anuglarProjectMapping as Mapping)
+
+  const angularPageGenerator = createAngularComponentGenerator()
+
+  const angulaRootModuleGenerator = createComponentGenerator()
```
alexnm: angular

packages/teleport-project-generator-angular/src/index.ts
```diff
@@ -0,0 +1,98 @@
...
+import { createPlugin as createImportPlugin } from '@teleporthq/teleport-plugin-import-statements'
+
+import { CUSTOM_BODY_CONTENT } from './constants'
+import anuglarProjectMapping from './angular-mapping.json'
+
+export const createPreactProjectGenerator = () => {
+  const prettierJS = createPostProcessor({ fileType: FILE_TYPE.TS })
+  const importStatementsPlugin = createImportPlugin({ fileType: FILE_TYPE.TS })
+
+  const rootModuleGeneratorAngular = createAngularModuleGenerator({ moduleType: 'root' })
```
alexnm: maybe you can drop the Angular in the name here, it feels a bit too verbose

packages/teleport-types/src/generators.ts
```diff
@@ -154,6 +162,13 @@ export interface ProjectStrategy {
   }
 }
 
+export interface CustomTag {
```
alexnm: can we reuse the code from CustomLinkTag? or use this for CustomLinkTag so we don't duplicate?

packages/teleport-types/src/generators.ts
```diff
@@ -111,9 +112,12 @@ export interface ProjectStrategy {
     path: string[]
     options?: {
       createFolderForEachComponent?: boolean
-      customComponentFileName?: string // only used when createFolderForEachComponent is true
-      customStyleFileName?: string
-      customTemplateFileName?: string
+      customComponentFileName?: (name?: string) => string // only used when createFolderForEachComponent is true
+      customStyleFileName?: (name?: string) => string
+      customTemplateFileName?: (name?: string) => string
+      module?: {
```
alexnm: I would name this `moduleGenerator` and place it next to `generator` and `path` since it's an important part of the  strategy

packages/teleport-project-generator-angular/src/index.ts
```diff
@@ -0,0 +1,98 @@
+import { createPlugin as createAngularModuleGenerator } from '@teleporthq/teleport-plugin-module-generator-angular'
```
alexnm: is this a plugin or a generator?

packages/teleport-project-generator-angular/src/index.ts
```diff
@@ -0,0 +1,98 @@
...
+      fileName: 'index',
+      options: {
+        appRootOverride: CUSTOM_BODY_CONTENT,
+        customTags: [
+          { tagName: 'base', attributeKey: 'href', attributeValue: '/', targetTag: 'head' },
+        ],
+      },
+    },
+    static: {
+      prefix: '/assets',
```
alexnm: I adjusted the static values in all the strategies in #355, have a look

packages/teleport-shared/src/node-handlers/node-to-html/index.ts
```diff
@@ -22,7 +22,7 @@ const generateElementNode: NodeToHTML<UIDLElementNode, HastNode> = (node, params
     ? createHTMLNode(templateSyntax.customElementTagName(camelCaseToDashCase(elementType)))
     : createHTMLNode(elementType)
 
-  if (dependency) {
+  if (dependency && !templateSyntax.shouldOmitLocalDependencies) {
```
alexnm: you can use the same flag as the node-to-jsx, `dependencyHandling`: 'import' | 'ignore'. then add the condition in here

packages/teleport-shared/src/node-handlers/node-to-html/types.ts
```diff
@@ -21,6 +21,7 @@ export interface HTMLTemplateSyntax {
   repeatAttr?: string
   repeatIterator?: (iteratorName: string, iteratedCollection: string, useIndex: boolean) => string
   customElementTagName?: (value: string) => string
+  shouldOmitLocalDependencies?: boolean
```
alexnm: see my comment from node-to-html/index.ts

packages/teleport-project-generator/src/file-handlers.ts
```diff
@@ -43,6 +47,40 @@ export const createComponent = async (
   return strategy.components.generator.generateComponent(componentUIDL, options)
 }
 
+export const createComponentModule = async (uidl: ProjectUIDL, strategy: ProjectStrategy) => {
```
alexnm: maybe createModule?

packages/teleport-project-generator/src/index.ts
```diff
@@ -89,6 +91,15 @@ export const createProjectGenerator = (strategy: ProjectStrategy): ProjectGenera
 
       injectFilesToPath(rootFolder, path, files)
       collectedDependencies = { ...collectedDependencies, ...dependencies }
+
+      const { module } = strategy.pages.options || {
```
alexnm: this would be simplified if module would be placed next to generator, not inside options

packages/teleport-types/src/generators.ts
```diff
@@ -70,6 +70,7 @@ export interface GeneratorOptions {
   skipValidation?: boolean
   projectRouteDefinition?: UIDLStateDefinition
   strategy?: ProjectStrategy
+  componentsList?: string[]
```
alexnm: maybe this could be renamed as `moduleComponents`

packages/teleport-project-generator/src/file-handlers.ts
JayaKrishnaNamburu: We need to generate three different kinds of modules for Angular, one for root and second for pages and for components, just to differentiate the three, I named them like this. How about adding a comment on the top?

alexnm: ah, got it, makes sense

packages/teleport-plugin-module-generator-angular/src/index.ts
JayaKrishnaNamburu: I thought i am the only one, who does this 😂

packages/teleport-project-generator/src/index.ts
JayaKrishnaNamburu: Yeah makes sense, but all projects don't need this. Since it is a flavor specific one, placed it under options

alexnm: yes, but it can be optional and not used by other projects.

packages/teleport-project-generator-angular/src/index.ts
JayaKrishnaNamburu: Yes, Angular in too many places is causing confusion too 😅

packages/teleport-project-generator-angular/src/index.ts
JayaKrishnaNamburu: plugin, let me remove generator to elimintate the confusion

packages/teleport-plugin-module-generator-angular/src/utils.ts
JayaKrishnaNamburu: No, as by default we push all the components into the components folder according to the folder structure.

packages/teleport-types/src/generators.ts
JayaKrishnaNamburu: Yes, had it in my list. We can merge all three and make one

JayaKrishnaNamburu:
Angular's project template needs some change. It breaks when we run `yarn standalone`. Here is the template --> [packages/teleport-project-generator-angular/src/project-template.ts](https://github.com/teleporthq/teleport-code-generators/blob/9620e4abee8d671f22612bd3a2a2ad92a2e64e0b/packages/teleport-project-generator-angular/src/project-template.ts)

JayaKrishnaNamburu:
> Angular's project template needs some change. It breaks when we run `yarn standalone`. Here is the template --> [packages/teleport-project-generator-angular/src/project-template.ts](https://github.com/teleporthq/teleport-code-generators/blob/9620e4abee8d671f22612bd3a2a2ad92a2e64e0b/packages/teleport-project-generator-angular/src/project-template.ts)

Fixed this issue 👍

JayaKrishnaNamburu:
fixes #360

packages/teleport-plugin-angular-base-component/src/index.ts
```diff
@@ -65,10 +66,11 @@ export const createPlugin: ComponentPluginFactory<AngularPluginConfig> = (config
         conditionalAttr: '*ngIf',
         repeatAttr: '*ngFor',
         repeatIterator: (iteratorName, iteratedCollection, useIndex) => {
-          const index = useIndex ? `; index as i` : ''
+          const index = useIndex ? `; index as index` : ''
```
JayaKrishnaNamburu: fixes #360

PR #365, opened 2019-09-18
feat(angular-base): changing title for tab dynamically

JayaKrishnaNamburu:


packages/teleport-plugin-angular-base-component/src/utils.ts
```diff
@@ -63,25 +66,31 @@ export const generateExportAST = (
...
   )
 }
 
-const constructorAST = (t = types) => {
-  return t.classMethod('constructor', t.identifier('constructor'), [], t.blockStatement([]))
+const constructorAST = (seo: UIDLComponentSEO, t = types) => {
+  const params = []
+  if (seo) {
+    const seoBinding = t.identifier(`private titleService: Title`)
+    params.push(seoBinding)
```
JayaKrishnaNamburu: These are *typescript* specific syntaxes. Since we are using babel for constructing our AST, in most cases it works. But in cases like reserved words like `private` `public` can't be generated. Since babel parser expected only identifier in the position.

Even in the case of export class with Decorator, we were able to generate because we are taking advantage of babel's support for Flow and Typescript. Like we are combining a Flow syntax AST and normal JS AST and linking the chunks together, so the export decorator is able.

packages/teleport-postprocessor-prettier-js/src/index.ts
```diff
@@ -15,12 +16,16 @@ export const createPostProcessor = (options: PostProcessorFactoryOptions = {}) =
   const fileType = options.fileType || FILE_TYPE.JS
   const formatOptions = { ...PRETTIER_CONFIG, ...options.formatOptions }
 
+  const jsParser = fileType === FILE_TYPE.TS ? parserTypescript : parserBabylon
+  const plugins = [jsParser, parserPostCSS]
+  const parser = fileType === FILE_TYPE.TS ? 'typescript' : 'babel'
```
JayaKrishnaNamburu: As I am trying a workaround here --> https://github.com/teleporthq/teleport-code-generators/pull/365/files#diff-cdb7f45628daba09fcfd879785f861e7R89 for adding reserved keywords.

Again prettier throws an error in the post-processing phase. Since prettier is currently using `babel-parser` for all TS files too. 

But this breaks when we try to do the same with more *typescript* specific syntaxes. So we need to use target-specific parsers to prettify the resultant code.

JayaKrishnaNamburu: Yeah, in a way that too makes sense. As we can just use either *prettier-ts* or *prettier-js* depending on the project. Will make those changes.

Issue #292, opened 2019-06-27
HTML formatting utility


alexnm:
We're relying on `prettier` for formatting html chunks (which are represented as strings). However, perf tests (#64) showed that prettier is running very slow when the UIDL gets bigger (ex: 2000 nodes).

We have two ways of approaching this:

### Simple post-processor function that formats html tags
This should run after the chunks are linked and should just align the html tags and content accordingly. Important to note that some frameworks might have non-standard html elements to render.

### HAST formatter
This should run at link time, when the html is a HAST structure represented internally. some utilities from the unified ecosystem could be checked out for this.

alexnm:
we're keeping prettier-html for now, it's still the best solution we have

PR #370, opened 2019-09-24
feat(js-beautifier): using js-beautify for formatting html

JayaKrishnaNamburu:
Discussion regarding #292

### DIFF
--- a/packages/teleport-postprocessor-js-beautify-html/src/index.ts
+++ b/packages/teleport-postprocessor-js-beautify-html/src/index.ts
@@ -1,0 +1,32 @@
+// @ts-ignore
+import { html as beautifyHTML } from 'js-beautify'
+
+import { PostProcessor, PrettierFormatOptions, FileType } from '@teleporthq/teleport-types'
+
+interface PostProcessorFactoryOptions {
+  fileType?: string
+  formatOptions?: PrettierFormatOptions
+}
+
+export const createPostProcessor = (options: PostProcessorFactoryOptions = {}) => {
+  const fileType = options.fileType || FileType.HTML
+
+  const processor: PostProcessor = (codeChunks) => {
+    if (codeChunks[fileType]) {
+      const result = beautifyHTML(codeChunks[fileType], {
+        wrap_line_length: 80,
+        end_with_newline: true,
+        preserve_newlines: true,
+      })
+      codeChunks[fileType] = result
+    } else {
+      console.warn('No code chunk of type HTML found, prettier-html did not perform any operation')
+    }
+
+    return codeChunks
+  }
+
+  return processor
+}
+
+export default createPostProcessor()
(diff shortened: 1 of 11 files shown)


packages/teleport-postprocessor-js-beautify-html/src/index.ts
```diff
@@ -0,0 +1,32 @@
+// @ts-ignore
```
alexnm: let's check if @types/js-beautify exists so we can have typing support for all dependencies

packages/teleport-postprocessor-js-beautify-html/src/index.ts
```diff
@@ -0,0 +1,32 @@
...
+  formatOptions?: PrettierFormatOptions
+}
+
+export const createPostProcessor = (options: PostProcessorFactoryOptions = {}) => {
+  const fileType = options.fileType || FileType.HTML
+
+  const processor: PostProcessor = (codeChunks) => {
+    if (codeChunks[fileType]) {
+      const result = beautifyHTML(codeChunks[fileType], {
+        wrap_line_length: 80,
```
alexnm: can we extract this into constants? same as prettier settings

packages/teleport-postprocessor-js-beautify-html/src/index.ts
```diff
@@ -0,0 +1,32 @@
...
+        end_with_newline: true,
+        preserve_newlines: true,
+      })
+      codeChunks[fileType] = result
+    } else {
+      console.warn('No code chunk of type HTML found, prettier-html did not perform any operation')
```
alexnm: make sure you rename the message here

packages/teleport-postprocessor-js-beautify-html/src/index.ts
JayaKrishnaNamburu: yup 👍

packages/teleport-postprocessor-js-beautify-html/src/index.ts
JayaKrishnaNamburu: Yeah yeah 😅  this was for testing purpose. I just copied 😅

JayaKrishnaNamburu:
Hey @alexnm i noticed this --> https://github.com/teleporthq/teleport-code-generators/blob/master/packages/teleport-postprocessor-prettier-js/src/index.ts#L21

Here is my opinion, currently we are using prettier for formatting and so we got the flexibility to parse js and jsx content both with babel. But when we are trying to add more formatters, I noticed this behavior.

1. A pure JS Formatter can't format a JSX file (does not break, but formats it very poorly)
2. For example, let us take the behavior of *project-generator-react* the output file is *js* but actually it has `jsx` inside it. Which babel takes care out of the box but other pure JS formatters tend to fail. 
- I feel we need to have something like a content flag, which describes the type of content inside the file. So then we can do something like this --> https://github.com/teleporthq/teleport-code-generators/pull/370/files#diff-3a9f5f76c2ea1b8e9388bb72e21959d6R7

alexnm:
closing this as it doesn't work out with what we need

PR #382, opened 2019-10-13
feat(teleport-component-generator): added template for preact codesan…

### LINKED ISSUE #367: Investigate the preact codesandbox export
https://codesandbox.io/s/src6v

At first it seems that the error is caused by .babelrc, as babel.config.js doesn't throw an error. However, that still doesn't work as it outputs a blank screen.

Also, we might look into preact-helmet, it seems that react-helmet doesn't work with this setup without preact-compact.

However, locally the project builds just fine, so maybe it's a thing with preact-cli?

JayaKrishnaNamburu:
…dbox

fixes #367

### DIFF
--- a/packages/teleport-code-generator/src/types.ts
+++ b/packages/teleport-code-generator/src/types.ts
@@ -49,7 +49,6 @@ export enum ProjectType {
   PREACT = 'Preact',
   STENCIL = 'Stencil',
   ANGULAR = 'Angular',
-  PREACT_CODESANDBOX = 'PreactCodesandbox',
 }
 
 export enum ComponentType {
--- a/packages/teleport-code-generator/src/types.ts
+++ b/packages/teleport-code-generator/src/types.ts
@@ -49,6 +49,7 @@ export enum ProjectType {
   PREACT = 'Preact',
   STENCIL = 'Stencil',
   ANGULAR = 'Angular',
+  PREACT_CODESANDBOX = 'PreactCodesandbox',
 }
 
 export enum ComponentType {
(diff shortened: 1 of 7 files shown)


packages/teleport-code-generator/src/types.ts
```diff
@@ -49,6 +49,7 @@ export enum ProjectType {
   PREACT = 'Preact',
   STENCIL = 'Stencil',
   ANGULAR = 'Angular',
+  PREACT_CODESANDBOX = 'PreactCodesandbox',
```
alexnm: can we keep this out of ProjectType? and just decide on the template based on the publisher type? ProjectType is used by other projects to display all available outputs

packages/teleport-test/src/standalone.ts
```diff
@@ -49,6 +49,11 @@ const run = async () => {
     console.info(ProjectType.PREACT, '-', result.payload)
     result = await packProject(projectUIDL, { ...packerOptions, projectType: ProjectType.ANGULAR })
     console.info(ProjectType.ANGULAR, '-', result.payload)
+    result = await packProject(projectUIDL, {
```
alexnm: can the packProject function decide which template to use internally? based on the publisher

packages/teleport-code-generator/src/types.ts
JayaKrishnaNamburu: Yeah, feels good that way. But I have a doubt if the user tries to download which code we should let them to? Since we will have two different codebases.

alexnm: let's use the default one and the codesandbox one is specific for the codesandbox publisher

PR #384, opened 2019-10-14
feat(proj-gen-gatsby): support for project generators using gatsby

JayaKrishnaNamburu:
- [x]  Support for a custom HTML file


JayaKrishnaNamburu:
I feel the Gatsby community is more leaned towards the camelCase for css Modules. And after checking the current stage in which the generators are, I felt its good to switch to `camelCase` by default. So, in cases like preact we can configure them in their project generators when building.

JayaKrishnaNamburu:
Currently, this creates projects powered by gatsby, and the custom Html can be controlled. Regarding `manifest` i have few ideas as we can consider them as framework-specific changes. Will make them in a separate PR

packages/teleport-project-generator-gatsby/README.md
```diff
@@ -0,0 +1,14 @@
+# teleport-project-generator-gatsby
+
+Project generator using React and React-Router, on top of a standard webpack config.
```
alexnm: update description

packages/teleport-project-generator-gatsby/__tests__/end2end/index.ts
```diff
@@ -0,0 +1,38 @@
+// @ts-ignore
```
alexnm: you don't need the @ts-ignore for the json files anymore

packages/teleport-project-generator-gatsby/src/gatsby-mapping.json
```diff
@@ -0,0 +1,18 @@
+{
+  "elements": {
+    "navlink": {
+      "elementType": "Link",
+      "dependency": {
+        "type": "library",
+        "path": "gatsby",
+        "version": "^2.15.36",
```
alexnm: no need to specify the version here if we have it in package.json

PR #386, opened 2019-10-15
Feature/gatsby styled components

JayaKrishnaNamburu:


JayaKrishnaNamburu:
Fixing the issue, which is causing to break the template. Here is the codesandbox generated with this setup --> https://codesandbox.io/s/ydrg8

PR #388, opened 2019-10-17
feat(gridsome): generating projects based on gridsome

JayaKrishnaNamburu:


packages/teleport-project-generator-gridsome/src/gridsome-project-mapping.json
```diff
@@ -0,0 +1,10 @@
+{
+  "elements": {
+    "navlink": {
+      "elementType": "router-link",
```
alexnm: is this the same mapping as vue? if so it can be skipped

JayaKrishnaNamburu: Yeah it's the same mapping 👍

PR #395, opened 2019-10-24
fix(vue-base): making the vue component imports html friendly

### LINKED ISSUE #394: Append prefix for the component names in Vue
Currently, in Vue, we are just importing the component and not appending any prefixes while using them. This works fine most of the time unless the name of the component collides with an HTML tag.

In Angular we append app so it generates

``<app-header></app-header>``

but in Vue it generates

``<header></header>`` which is works as an HTML tag instead of the component.

JayaKrishnaNamburu:
fixes #394

### DIFF
--- a/packages/teleport-plugin-vue-base-component/src/index.ts
+++ b/packages/teleport-plugin-vue-base-component/src/index.ts
@@ -6,6 +6,7 @@ import {
   ChunkType,
 } from '@teleporthq/teleport-types'
 import { createHTMLTemplateSyntax } from '@teleporthq/teleport-plugin-common'
+import { UIDLUtils } from '@teleporthq/teleport-shared'
 
 import {
   DEFAULT_VUE_TEMPLATE_CHUNK_NAME,
@@ -52,7 +53,12 @@ export const createVueComponentPlugin: ComponentPluginFactory<VueComponentConfig
           const iterator = useIndex ? `(${iteratorName}, index)` : iteratorName
           return `${iterator} in ${iteratedCollection}`
         },
-        customElementTagName: (value) => value,
+        customElementTagName: (value) => {
+          return {
+            tagName: UIDLUtils.createWebComponentFriendlyName(value),
+            importName: UIDLUtils.createWebComponentFriendlyImportName(value),
+          }
+        },
       }
     )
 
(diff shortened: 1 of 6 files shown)


packages/teleport-plugin-vue-base-component/src/index.ts
```diff
@@ -52,7 +53,8 @@ export const createVueComponentPlugin: ComponentPluginFactory<VueComponentConfig
           const iterator = useIndex ? `(${iteratorName}, index)` : iteratorName
           return `${iterator} in ${iteratedCollection}`
         },
-        customElementTagName: (value) => value,
+        customElementTagName: (value) => UIDLUtils.createWebComponentFriendlyName(value),
+        customElementImportName: (value) => `App${value}`,
```
alexnm: I wouldn't create a separate import name as the two need to be in sync to work

alexnm: what I mean is that you can't have two different functions, the difference between the tag name and the import name is the camel case or dash case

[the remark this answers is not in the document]
JayaKrishnaNamburu: Yeah, noticed these cases now. Checking those

JayaKrishnaNamburu:
Tests are breaking because, the dependencies are mentioned as local in the tests. If we change that, they will be all right. @alexnm please check

packages/teleport-plugin-vue-base-component/src/index.ts
```diff
@@ -52,7 +53,12 @@ export const createVueComponentPlugin: ComponentPluginFactory<VueComponentConfig
           const iterator = useIndex ? `(${iteratorName}, index)` : iteratorName
           return `${iterator} in ${iteratedCollection}`
         },
-        customElementTagName: (value) => value,
+        customElementTagName: (value) => {
+          return {
+            tagName: UIDLUtils.createWebComponentFriendlyName(value),
+            importName: UIDLUtils.createWebComponentFriendlyImportName(value),
```
alexnm: can't importName be: dashToCamelCase(tagname)?

alexnm:
closing this as it was fixed on master

PR #398, opened 2019-10-25
feat(gatsby-proj): support for custom manifest in gatsby

### LINKED ISSUE #397: Custom manifest support for gatsby and gridom
Configure both the templates to use the manifest that is generated from code generators.

JayaKrishnaNamburu:
fixes #397 

I cross-checked and we don't need to make changes for gridsome.


packages/teleport-project-generator-gatsby/src/utils.ts
```diff
@@ -69,6 +69,11 @@ const generateHTMLNode = (uidl: ProjectUIDL, options: EntryFileOptions, t = type
   const headNode = ASTBuilders.createJSXTag('head')
   const bodyNode = ASTBuilders.createJSXTag('body')
   const noScriptNode = ASTBuilders.createJSXTag('noscript')
+  const manifestTag = ASTBuilders.createJSXTag('link')
```
alexnm: I think you should check here if uidl.globals.manifest exists, otherwise it doesn't make sense to add the manifest link

JayaKrishnaNamburu: yes, should add that 👍

PR #399, opened 2019-10-29
feat(gatsby-proj): project generators can configure a file to change

JayaKrishnaNamburu:
A Different approach for PR --> https://github.com/teleporthq/teleport-code-generators/pull/386 using `magic-string` for reducing bundle size

### DIFF
--- a/packages/teleport-project-generator/src/index.ts
+++ b/packages/teleport-project-generator/src/index.ts
@@ -31,6 +31,7 @@ import {
   ComponentGenerator,
   ProjectStrategyPageOptions,
 } from '@teleporthq/teleport-types'
+import MagicString from 'magic-string'
 
 type UpdateGeneratorCallback = (generator: ComponentGenerator) => void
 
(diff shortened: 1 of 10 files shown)


JayaKrishnaNamburu:
IMO, this approach is much smaller in bundle size and complexity. But one change i would like to see is moving the logic from `teleport-project-generator` to `teleport-project-generator-gatsby` since it is a gatsby specific change and not needed for other projects.

packages/teleport-project-generator/src/index.ts
```diff
@@ -192,6 +195,30 @@ export class ProjectGenerator {
       injectFilesToPath(rootFolder, this.strategy.components.path, [componentsModuleFile])
     }
 
+    // Handling framework specific changes to the project
+    const { framework } = this.strategy
+
+    if (framework && framework.config) {
+      const { fileName, fileType, styleVariation } = framework.config
```
alexnm: just a thought when seeing this, can we maybe think of a way to create plugins for project generators? these plugins would do these kind of string changes, config updates, etc.

JayaKrishnaNamburu: Yeah, sounds interesting

PR #415, opened 2019-11-26
fix(styled-comp): appending prefix when name clashes 

### LINKED ISSUE #412: Potential naming conflict with StyledComponents
When the styled components tags are generated (eg: const Navbar = styled.nav) based on the names of the elements, the tags can be in naming conflict with the component itself.

One quick solution could be to check for that particular naming conflict and add a `Styled` prefix for that tag

JayaKrishnaNamburu:
Appending ``Styled`` as a prefix when ``className`` is clashing with the component name or any of its dependencies

fixes #412


JayaKrishnaNamburu:
Hey, looks like this commit came from my fork. I will switch from next PR's, recently I switched to my fork to work on integration with code sandbox.

PR #439, opened 2020-03-23
feat(link) adding basic support for links on any nodes and resolver functionality

alexnm:
Adding support for the link sub-structure for each element node, which is transformed at resolver time into a link/navlink element wrapping the corresponding node

### DIFF
--- a/examples/uidl-samples/project.json
+++ b/examples/uidl-samples/project.json
@@ -171,7 +171,7 @@
                                   "abilities": {
                                     "link": {
                                       "type": "section",
-                                      "options": {
+                                      "content": {
                                         "id": "test"
                                       }
                                     }
@@ -415,7 +415,7 @@
                           "abilities": {
                             "link": {
                               "type": "navlink",
-                              "options": {
+                              "content": {
                                 "routeName": "home"
                               }
                             }
@@ -454,7 +454,7 @@
                           "abilities": {
                             "link": {
                               "type": "navlink",
-                              "options": {
+                              "content": {
                                 "routeName": "team"
                               }
                             }
@@ -493,7 +493,7 @@
                           "abilities": {
                             "link": {
                               "type": "navlink",
-                              "options": {
+                              "content": {
                                 "routeName": "contact-us"
                               }
                             }
@@ -553,7 +553,7 @@
                 "abilities": {
                   "link": {
                     "type": "url",
-                    "options": {
+                    "content": {
                       "url": "$props.link",
                       "newTab": true
                     }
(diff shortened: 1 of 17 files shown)


packages/teleport-types/src/uidl.ts
```diff
@@ -189,6 +192,23 @@ export type UIDLStyleDefinitions = Record<string, UIDLStyleValue>
...
 export type UIDLEventDefinitions = Record<string, UIDLEventHandlerStatement[]>
 
+export type UIDLLinkURLDefinition = {
+  type: 'url'
+  options: { url: UIDLAttributeValue; newTab: boolean }
+}
+export type UIDLLinkPageDefinition = { type: 'navlink'; options: { routeName: string } }
+export type UIDLLinkMailDefinition = {
+  type: 'mail'
+  options: { mail: string; subject?: string }
```
JayaKrishnaNamburu: Hey, can we consider adding support for **body** too on the code-generators side alone. Down the line, we can just add it when needed on the UI.

packages/teleport-uidl-validator/src/uidl-schemas/component.json
```diff
@@ -299,6 +299,12 @@
         "key": {"type": "string"},
         "dependency": {"$ref": "#/definitions/componentDependency"},
         "style": {"$ref": "#/definitions/styleDefinitions"},
+        "abilities": {
+          "type": "object",
+          "properties": {
+            "link": {"$ref": "#/definitions/linkDefinition"}
+          }
+        },
```
JayaKrishnaNamburu: TODO: Update project-schema at `https://docs.teleporthq.io/uidl-schema/v1/project.json` to support the new custom-code and abilities. UIDL's downloaded from playground still point to it. Nice to have both in sync :)

examples/uidl-samples/project.json
```diff
@@ -526,15 +547,22 @@
...
               "type": "element",
               "content": {
                 "elementType": "image",
-                "attrs": {
-                  "url": "$props.url"
+                "abilities": {
+                  "link": {
+                    "type": "url",
+                    "options": {
+                      "url": "$props.link",
```
JayaKrishnaNamburu: Update validator to check for the url and link in abilities if they are being used from props. Right now if we run it throws validation errors for `link` and `url` even if they are being used. This component should throw only `url` from `propDefinitions` in validations.

alexnm: I added link.options.url in the traverse nodes function, it should be passed now when the validator is looking at all usages of UIDLAttributes. can you check again?

PR #459, opened 2020-07-16
feat(proj-styles): adding support for conditions for project-styles

JayaKrishnaNamburu:
## Enhancements 
We are now allowing to add conditions for the project styles. In previous release, we allowed support for project-styles and made them available to be added to any nodes.
Now, we are allowing conditions on top of project-style. Which allows the user to design the style once in all the conditions like, **default**, **media**, **pseudo-state**. So when a node is referred with the style, it inherits the behaviour in all the conditions.

## Refactors

- Adding links to elements like button's is a commonly used practice but **semantically-incorrect**. So, by default we are converting them to use `a` tags with `button-styles` from the themes.
- By default Project theme is not added to head anymore, we just convert them as `project-styles` from playground. So we could refer them on elements as styles directly.

### DIFF
--- a/packages/teleport-uidl-resolver/src/resolvers/abilities/utils.ts
+++ b/packages/teleport-uidl-resolver/src/resolvers/abilities/utils.ts
@@ -43,8 +43,21 @@ export const insertLinks = (
       return node
     }
 
-    // a text node (span) on which we added a link gets transformed into an <a>
-    // the rest of the text elements get wrapped with an <a> tag
+    /* We repalce buttons with link to use <a> tag's, to make the generated
+    code to be semantically correct. */
+    if (elementType === 'button') {
+      node.content.elementType = getLinkElementType(abilities.link)
+      node.content.semanticType = ''
+      node.content.attrs = createLinkAttributes(abilities.link, options)
+      node.content.style = {
+        ...node.content.style,
+        textDecoration: { type: 'static', content: 'none' },
+      }
+      return node
+    }
+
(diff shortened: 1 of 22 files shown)


packages/teleport-uidl-resolver/src/resolvers/abilities/utils.ts
```diff
@@ -43,8 +43,21 @@ export const insertLinks = (
...
-    // the rest of the text elements get wrapped with an <a> tag
+    /* We repalce buttons with link to use <a> tag's, to make the generated
+    code to be semantically correct. */
+    if (elementType === 'button') {
+      node.content.elementType = getLinkElementType(abilities.link)
+      node.content.semanticType = ''
+      node.content.attrs = createLinkAttributes(abilities.link, options)
+      node.content.style = {
+        ...node.content.style,
+        textDecoration: { type: 'static', content: 'none' },
```
JayaKrishnaNamburu: @alexnm what are your ideas on this, should we overwrite these behaviours including, active, hover and visited too. Since they are not rendered in stage as they are basically buttons and causes inconsistencies when published !!

alexnm: not sure if this is needed (the textDecoration). when a link is set, the node inherits the style from the link style from the theme (eg: text-decoration or color if they are set). if the text in the button appears underlined, then the user should reset the underline from the inspector, hence it will appear in the UIDL.

alexnm (approved):

PR #467, opened 2020-07-30
ci(package.json): add stream flag for build script to reduce time for build time

JayaKrishnaNamburu:


JayaKrishnaNamburu:
Looks like it's taking the same time, with or without the stream flag

PR #471, opened 2020-08-05
refactor(teleport-types): refactored checkes and validators for UIDLDependency

### LINKED ISSUE #470: External library dependencies are missing from final package.json file
## Problem

When a node is using any external dependency, example dependency type as `library`. These dependencies are missing from the final `package.json` file.

Here is an example node which uses external library
```js
{
  "type": "element",
  "content": {
    "elementType": "component",
    "semanticType": "Button",
    "children": [],
    "dependency": {
      "type": "library",
      "path": "antd",
      "meta": {
        "namedImport": true
      }
    },
    "style": {},
    "attrs": {
      "type": {
        "type": "static",
        "content": "danger"
      }
    },
    "abilities": {},
    "referencedStyles": {}
  }
}

```

When using a component generator on top of this, we get the following code
```js
import { Button } from 'antd'
```

But the final `package.json` is missing these external dependencies. Here is the `package.json` at the end.

```js
"dependencies": {
    "react": "^16.12.0",
    "react-dom": "^16.12.0",
    "react-router-dom": "5.1.2",
    "react-helmet": "^5.2.1",
    "react-scripts": "3.2.0"
  },
```

**antd** is missing from the project's `package.json` file.

Codesandbox url --> https://codesandbox.io/s/lvs5u

## Solution

**antd** should be added to the `package.json` at the end if it is being used.

JayaKrishnaNamburu:
fixes #470 

Currently the type of UIDLDependency has `version` as optional. But when we are using a `package` or `library` the `version` is mandatory. Right now it's passing the checks since it's an optional field. Just refactored the checkes.

### DIFF
--- a/packages/teleport-plugin-vue-app-routing/src/index.ts
+++ b/packages/teleport-plugin-vue-app-routing/src/index.ts
@@ -22,14 +22,17 @@ export const createVueAppRoutingPlugin: ComponentPluginFactory<VueRouterConfig> 
     dependencies.Vue = {
       type: 'library',
       path: 'vue',
+      version: '^2.6.7',
     }
     dependencies.Router = {
       type: 'library',
       path: 'vue-router',
+      version: '^3.0.2',
     }
     dependencies.Meta = {
       type: 'library',
       path: 'vue-meta',
+      version: '^2.2.1',
     }
 
     const routerDeclaration = t.expressionStatement(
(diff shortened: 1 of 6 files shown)


packages/teleport-plugin-vue-app-routing/src/index.ts
```diff
@@ -22,14 +22,17 @@ export const createVueAppRoutingPlugin: ComponentPluginFactory<VueRouterConfig>
     dependencies.Vue = {
       type: 'library',
       path: 'vue',
+      version: '^2.6.7',
```
JayaKrishnaNamburu, line 25: Previously all these values are kind of being picked by the templates. If the templates don't have these dependencies specified. Then then these dependencies won't reflect in `package.json` even if we are using them here.

alexnm (approved):

PR #478, opened 2020-08-13
feat(core): importing modules or packages and using them as reference for nodes

### LINKED ISSUE #465: Adding imports as a reference in ComponentUIDL
## Current Behaviour

1. Right now we can specify `node` level dependency and `project` level dependency.
https://github.com/teleporthq/teleport-code-generators/blob/development/packages/teleport-types/src/uidl.ts#L171
If, we consider a UIDLElement
```js
export interface UIDLElement {
  elementType: string
  semanticType?: string
  name?: string
  key?: string // internal usage
  dependency?: UIDLDependency
  style?: UIDLStyleDefinitions
  attrs?: Record<string, UIDLAttributeValue>
  events?: UIDLEventDefinitions
  abilities?: {
    link?: UIDLLinkNode
    // In the future more element abilities can be added here
  }
  referencedStyles?: UIDLReferencedStyles
  children?: UIDLNode[]
  selfClosing?: boolean
  ignore?: boolean
}
```

At generation time, we parse and the node and add the specified dependency to dependency list. Which works great, but when we are looking at generating components which are using external libraries. We might some times just add a couple of `import` statements. Or just `import` and refer as a attribute for a element. RIght now, we don't have that flexibility in the generators and UIDL.

```jsx
import { ThemeProvider } from 'react-ui'
import { tokens, components } from 'react-ui/themes/light'

ReactDOM.render(  
<ThemeProvider tokens={tokens} components={components}>  
 <App />  
</ThemeProvider>,  
document.getElementById('root'))
```

Let's take this pice of code as example. Here we are importing `tokens` and `components` and using them as just reference. But currently we don't have a way to provide this functionality in UIDL. It is going to be a good to have feature, where we can use repl alone to quickly write some components and see how they are rendering.

The closest we could come right now is --> https://teleport-repl-git-feature-preview-component.teleport-team.vercel.app/?uidlLink=e6yy8exwkrjurk4bsp61uc

## Proposal

The proposal is to have a `attribute` on `ComponentUIDL` and then allow the `attributes` to refer to them for imports

```js
export interface ComponentUIDL {
  $schema?: string
  name: string
  node: UIDLElementNode
  importDefinitions?: Record<string, UIDLDependency> // This is where we specify all the additional imports we need
  styleSetDefinitions?: Record<string, UIDLStyleSetDefinition>
  propDefinitions?: Record<string, UIDLPropDefinition>
  stateDefinitions?: Record<string, UIDLStateDefinition>
  outputOptions?: UIDLComponentOutputOptions
  seo?: UIDLComponentSEO
}
```

And add a new reference.

```js
export interface UIDLImportReference {
   type: "dynamic",
   content: {
      referenceType: "import",
      id: <import-id>
  }
}
```

### LINKED ISSUE #474: Re-map imports when two dependencies in the component have same import specifier but from different path
## Issue

Currently, whenever we mention
```js
{
  ...node,
  "dependency": {
    "type": "package",
    "path": "@material-ui/core",
    "meta": {
      "namedImport": true
    }
  }
}
```
It adds an import from the external package which is mentioned here from `@material-ui/core` but if we try to import the same from multiple packages like here --> https://repl.teleporthq.io/?uidlLink=vakzcgvvly9j10t5ezgd5
Only, the last one is over-written. Because of how we are currently handling the imports-map internally.
https://github.com/teleporthq/teleport-code-generators/blob/development/packages/teleport-plugin-common/src/node-handlers/node-to-jsx/index.ts#L78

We are taking `tagName` alone into consideration. We should be cross-checking with package destinations too.

## Solution

1. One way to handle this, is to fine tune the validators to throw error upfront. So, the users can check the issue and change it to use alias for the import as shown below.
```js
{
  ...node,
  "dependency": {
    "type": "package",
    "path": "@material-ui/core",
    "meta": {
      "namedImport": true,
      "originalName": "MaterialButton"
    }
  }
}
```
2. The other possible solution is to make the import maps aware of what `component` is being imported from which `package`, so it will not overwrite the imports.

JayaKrishnaNamburu:
fixes #478 #474 

TODO

- [x] Import and use as attribute reference
- [x] Import and use as a direct node.
- [x] Use case for just import alone (Used for importing external css libraries) (Ideally while importing from certain design systems, we don't need to use this import in every single file. So, this need to be added only on the *root* node level. Ideally where we define our router)
- [x] Fix dependencies that are specified in the *root* node doesn't reflect in `package.json` file. 
- [x] Remap elementNames if the dependency is coming from external Import and need to have a alias name. Related to #474 

Checks 

- [x] Check the support for JSX based Generators like React, Preact, Stencil
- [x] Check support for HTML based Generators like Vue, Angular

Tests

- [x] Update test cases for all the new features that are being added.

### DIFF
--- a/packages/teleport-plugin-angular-base-component/src/utils.ts
+++ b/packages/teleport-plugin-angular-base-component/src/utils.ts
@@ -66,9 +66,12 @@ export const generateExportAST = (
     )
   })
 
-  const referencedImportDeclerations = Object.keys(importDefinitions).map((importRef) => {
-    return t.classProperty(t.identifier(importRef), t.identifier(importRef))
-  })
+  const referencedImportDeclerations = Object.keys(importDefinitions).reduce((acc, importRef) => {
+    if (!importDefinitions[importRef]?.meta?.importJustPath) {
+      acc.push(t.classProperty(t.identifier(importRef), t.identifier(importRef)))
+    }
+    return acc
+  }, [])
 
   const classBodyAST = (componentUIDL: ComponentUIDL) => {
     return t.classBody([
--- a/packages/teleport-plugin-angular-base-component/src/utils.ts
+++ b/packages/teleport-plugin-angular-base-component/src/utils.ts
@@ -24,6 +24,8 @@ export const generateExportAST = (
     angularMethodsAST = createMethodsObject(methodsObject, propDefinitions)
   }
 
+  const { importDefinitions = {} } = uidl
+
   const propDeclaration = Object.keys(propDefinitions).map((propKey) => {
     const definition = propDefinitions[propKey]
     // By default any prop with type function is used to emitting events to the callee
@@ -64,11 +66,16 @@ export const generateExportAST = (
     )
   })
 
+  const referencedImportDeclerations = Object.keys(importDefinitions).map((importRef) => {
+    return t.classProperty(t.identifier(importRef), t.identifier(importRef))
+  })
+
   const classBodyAST = (componentUIDL: ComponentUIDL) => {
     return t.classBody([
       ...propDeclaration,
       ...propertyDecleration,
       ...dataDeclaration,
+      ...referencedImportDeclerations,
       constructorAST(componentUIDL.seo),
       ...angularMethodsAST,
     ])
--- a/packages/teleport-plugin-angular-base-component/src/utils.ts
+++ b/packages/teleport-plugin-angular-base-component/src/utils.ts
@@ -66,7 +66,7 @@ export const generateExportAST = (
     )
   })
 
-  const referencedImportDeclerations = Object.keys(importDefinitions).reduce((acc, importRef) => {
+  const referencedImportDeclarations = Object.keys(importDefinitions).reduce((acc, importRef) => {
     if (!importDefinitions[importRef]?.meta?.importJustPath) {
       acc.push(t.classProperty(t.identifier(importRef), t.identifier(importRef)))
     }
@@ -78,7 +78,7 @@ export const generateExportAST = (
       ...propDeclaration,
       ...propertyDecleration,
       ...dataDeclaration,
-      ...referencedImportDeclerations,
+      ...referencedImportDeclarations,
       constructorAST(componentUIDL.seo),
       ...angularMethodsAST,
     ])
--- a/packages/teleport-test/src/standalone.ts
+++ b/packages/teleport-test/src/standalone.ts
@@ -1,12 +1,12 @@
-import { readFileSync } from 'fs'
+import { readFileSync, rmdirSync, mkdirSync } from 'fs'
 import { join } from 'path'
 import { packProject } from '@teleporthq/teleport-code-generator'
 import { ProjectUIDL, PackerOptions, ProjectType, PublisherType } from '@teleporthq/teleport-types'
 
 import reactProjectJSON from '../../../examples/uidl-samples/react-project.json'
-// import projectJSON from '../../../examples/uidl-samples/project.json'
+import projectJSON from '../../../examples/uidl-samples/project.json'
 
-// const projectUIDL = (projectJSON as unknown) as ProjectUIDL
+const projectUIDL = (projectJSON as unknown) as ProjectUIDL
 const reactProjectUIDL = (reactProjectJSON as unknown) as ProjectUIDL
 const assetFile = readFileSync(join(__dirname, 'asset.png'))
 const base64File = new Buffer(assetFile).toString('base64')
@@ -32,6 +32,9 @@ const packerOptions: PackerOptions = {
 
 const run = async () => {
   try {
+    rmdirSync('dist', { recursive: true })
+    mkdirSync('dist')
+
     let result
     // result = await packProject(projectUIDL, {
     //   ...packerOptions,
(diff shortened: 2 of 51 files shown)


JayaKrishnaNamburu:
Ensure all the frameworks support the basic requirements of using external components.

- [x] NextJS all it's `css` dependencies inside `_app` file instead of the `index` or router where we usually import

JayaKrishnaNamburu:
Things that are take care in this PR

## Bug Fixes

- Failling back to  `semanticType` and using it for imports when `elementType` is component.
Now if we have 2 imports with `dependency` type as `local` it will not generate as 

```js
import Component1 { Component2 } from './component'
```
Instead it will generate 

```js
import Component1 from './component1'
import Component2 from './component2'
```
Please look at the change here --> https://github.com/teleporthq/teleport-code-generators/pull/478/files#diff-eb4d5e0030567941fced2c15e0929989R94

- If you have two same components with different `dependencies`. Till now, the `dependency` which is defined at the end gets picked up. For example, if we are using `Button` from both `@chakra-ui/core` and `antd`. Only, one of them get's added to the import because of the import name clash. Now, it fixed and get's imported using import aliasing.

```js
import { ThemeProvider, Button, Avatar, tokens, components } from "react-ui";
import { Button as AntdButton } from "antd";
import {
  ThemeProvider as ChakraUicoreThemeProvider,
  Button as ChakraUicoreButton,
  theme
} from "@chakra-ui/core";
```
The imports are *aliased* automatically, so the users can write UIDL without bothering about import clashing.

## Features

- Introducing a new reference type called `import`
- Now, we can specify all the imports of a component using `importDefinitions` and the `id's` can be used for reference. To do so, you need to use `useAsReference` flag under `meta`. So, these are considered as imports used for reference and not an actual dependency.
- 
##### Defining importDefinitions

```js
 "importDefinitions": {
    "theme": {
      "type": "package",
      "path": "@chakra-ui/core",
      "version": "0.8.0",
      "meta": {
        "namedImport": true,
        "useAsReference": true
      }
    }
},
```

#####  Using as reference

```json
{
    ...element,
    attrs: {
        theme: {
            type: "dynamic",
            content: {
                referenceType: "import",
                id: "theme"
            }
        }
    }
}
```

- Import definitions can be directly as children.
```js
children: [
    {
        type: "dynamic",
        content: {
            referenceType: "import",
            id: "theme"
        }
    }
]
```
(PS: Don't know what could be real world use-case but it is supported now ;))

- And in most of the cases when we use external **packages** or **design systems** etc, we need to take care of `peerDependencies` which are needed for the present library to work. We don't usually use them within the application but we need them to reflect them in `package.json` and so they are resolved at install time. To solve this, now we have new attribute in `UIDL` that is `peerDefinitions`. These are used for defining the `package` and they are collected at the end and reflected in `package.json`

```js
"peerDefinitions": {
      "@emotion/core": {
        "type": "package",
        "path": "@emotion/core",
        "version": "^10.0.34"
      }
}
```

For example, if we take a UIDL like this --> https://repl.teleporthq.io/?uidlLink=xcw61tkbj0b1zv0hsto11kh and wrap it around a project, the result *codesandbox* will be something like this --> https://codesandbox.io/s/t3qhn?file=/src/index.js

### Project Generators

To add first-class support for external libraries, we need to make a few changes in **project-generators**. Since, every other framework has it's own style in handling few types of dependencies. For example, *NextJS* needs all the `css` based dependencies defined only in `_app.js` file. And *Angular* mostly imports in its *modules* and not in the components directly. This release mainly focuses on adding this support to *React* bases frameworks, since it's the straight forward at the moment.

#### NextJS

By default all `css` based dependencies in `nextjs` are removed and added to the `app.js` file. It is recomended to add these kind of global external `css` dependencies only in `importDefinitions` of `root` in `ProjectUIDL`

#### ReactJS

The CRA template doesn't need any such constraints, it works out of the box without any external configs.

#### VUEJS [Needs discussion]

Using external dependencies in VueJS based apps are not always straight forward. Some times we need to configure a couple of files to make them use it. But as of now, the basic import syntaxes work as other flavours. May be, we need independent plugins to handle such behaviours ?

#### Angular [Needs discussion]

Angular based apps, adds all the imports for things like external design system in their module. So, following the same any `element` using `external` dependency in `UIDL`. The `import` is added to the module instead. 

#### Stencil [Needs discussion]

I didn't come across any stencil apps using external dependencies, so stencil project-generator didn't undergo any change. As of now, stencil just ignore any dependencies. It might be `local` or `external`. The functionality is mostly aligned with **Angular**

JayaKrishnaNamburu:
- [ ] Dependency clash does still happen if we are referring a dependency with same name but from different packages. Need to be handled inside `node-to-jsx`

packages/teleport-plugin-angular-base-component/src/utils.ts
```diff
@@ -64,11 +66,19 @@ export const generateExportAST = (
     )
   })
 
+  const referencedImportDeclerations = Object.keys(importDefinitions).reduce((acc, importRef) => {
```
alexnm: typo Declarations

packages/teleport-test/src/standalone.ts
```diff
@@ -30,30 +32,50 @@ const packerOptions: PackerOptions = {
...
+      mkdirSync('dist')
+    }
+
     let result
     // result = await packProject(projectUIDL, {
     //   ...packerOptions,
     //   projectType: ProjectType.REACTNATIVE,
     // })
     // console.info(ProjectType.REACTNATIVE, '-', result.payload)
-    result = await packProject(projectUIDL, { ...packerOptions, projectType: ProjectType.REACT })
```
alexnm, line 39: you should revert the commenting here I guess

packages/teleport-uidl-resolver/src/utils.ts
```diff
@@ -440,7 +440,7 @@ const resolveDependency = (
 
     // ex: PrimaryButton component should be written in a file called primary-button
     // This is just a fallback for when the dependency path is not set by a project generator
-    const componentName = mappedElement.elementType
+    const componentName = mappedElement.semanticType ?? mappedElement.elementType
```
alexnm: are there situations where the mappedElement.semanticType exists? this is coming from the mapping files

alexnm (dismissed):

packages/teleport-uidl-resolver/src/utils.ts
JayaKrishnaNamburu: Yes in cases like custom mappings we need to consider this i feel, like a simple Button some external libraries like `chakra` in vue it seems to be `ca-button`. Not sure i am explaining correctly.

alexnm (approved):

PR #480, opened 2020-09-09
refactor(preact-template): upgrading preact to use preact x

### LINKED ISSUE #477: Upgrade preact template and the jsx syntax to use the latest hooks based syntaxes
Currently the `preact` component generator is still following the old `component` based code generation. We need to switch from component based generation to use `Functional Component` based hooks for all state management.

We support hooks based component generation for `react`, so it's jus that we need to switch `preact` to use it and  some dependency checks along with it.

JayaKrishnaNamburu:
Upgrading Preact to use Preaxt X and hooks which were recently released. So, we will be switching from using preact components to use hooks instead of class-components from now.

But there is a known issue with codesandbox. Codesandbox currently using `preact-compact` for aliasing instead of `preact/compact` for preact X. So, the sandboxed which are generated will not work.

https://github.com/codesandbox/codesandbox-client/issues/3685


JayaKrishnaNamburu:
Here are some series of tasks, that need to be done. Once the PR is merged

Update template for 

https://github.com/teleporthq/teleport-project-template-react

https://github.com/teleporthq/teleport-project-template-preact

https://github.com/teleporthq/teleport-project-template-preact-codesandbox

packages/teleport-plugin-jsx-head-config/src/index.ts
```diff
@@ -72,7 +72,7 @@ export const createJSXHeadConfigPlugin: ComponentPluginFactory<JSXHeadPluginConf
       rootElement.children.unshift(headConfigTag)
 
       dependencies[configTagIdentifier] = {
-        type: 'library',
+        type: 'package',
         path: configTagDependencyPath,
```
JayaKrishnaNamburu, line 76: `react-helmet` is a external package and need to be added to `package.json`. This should happen only if the package is used. Currently by default we added in templates which is not needed. And this is a requirement based usage and not a core part of react. So, we need to switch to `package` according to conventions.

packages/teleport-plugin-jsx-proptypes/src/index.ts
```diff
@@ -55,7 +55,7 @@ export const createPropTypesPlugin: ComponentPluginFactory<PropTypesConfig> = (c
     }
 
     dependencies.PropTypes = {
-      type: 'library',
+      type: 'package',
```
JayaKrishnaNamburu, line 58: same here

alexnm (dismissed):

alexnm (approved):

PR #491, opened 2020-10-19
refactor(html-mappings): add linebreak to html mappings

ionutpasca:


JayaKrishnaNamburu (approved):
Good catch, from the next version of playground we are going to reply mappings directly in th playground. Decoupling from the uidl-resolver.

PR #514, opened 2020-11-25
(feat): Plugins for project-generators 

### LINKED ISSUE #510: Ability to add "devDependencies" separately using project-generators
If we try to explore the path of #508  and allow users to write custom plugins. Then users should be able to inject `devDependencies` too which are not already included in the template. So, we need a way to allow users to add **devDependencies** along side with normal dependencies.

### LINKED ISSUE #522: RFC | Project Generator Plugins
We currently support plugins for project generators. But we don't have plugins to configure on project level.

These enable us to quickly change few files or rewrite few. Or just extend the default `project-generators` with few tweaks or adding new things. The initial discussion regarding the issue happened here #508

And the initial implementation for it is going on in #514 PR

JayaKrishnaNamburu:
- [x] Fix all the test cases
- [x] Add test cases for project-generators assembly line
- [x] Add test cases for individual project-plugins
- [x] Add integration test-cases for individual project-plugins


JayaKrishnaNamburu:
If we refer to this file --> https://github.com/teleporthq/teleport-code-generators/blob/feat/project-generator-plugins/packages/teleport-project-plugin-next-css-modules/src/index.ts

It is a test implementation of how the project-plugins can look like. One main issue is to handle the post-processors on top of project plugins.

Use case
-  The custom hand written code can still be formatted to match the code that is generated using code-generators.  

But currently, the post-processors will work using the chunking concept. So, we can't directly use them for formatting files. Should we need to change the current post-processors a little to solve this use case. Or have a completely new set of post-processors.

JayaKrishnaNamburu:
To address the above formatting concept. Having a project wide post-processor is not suggested. Since we have individual component formatter, running formatter again on top of all these files is not doing any good. 

Next, we need formatters to run on files that are being modified or custom generated. So we can directly format them using the existing formatter, which looks something like 

```js
const nextContent = prettierJS({
      [FileType.JS]: `const withCSS = require('@zeit/next-css')
    module.exports = withCSS({
      cssModules: true
    })`,
    })
```

We can discuss a little on how the API can be exposed etc. But with this, we are going to format the code using the same config that the generators use internally when they are used as default.

Here is how plugins will look like.

https://github.com/teleporthq/teleport-code-generators/pull/514/commits/952953e6a1079e7049ad5832066010ed0b3637d0#diff-3d9de7f13f3c7caa6a9c3b9623ceae18fe2e8a5130c3ba9dbf057fa2fccf495c

JayaKrishnaNamburu:
One DX with enabling plugins for project-generators is, we need to re-configure generators every time we want to switch just the style variation for the project.

### Problem

Currently we can configure style variation only for component generators. Something like

```js
const reactComponentGenerator = createReactComponentGenerator(
  ReactStyleVariation.CSSModules
);
reactComponentGenerator.addMapping(ReactProjectMapping);

const reactPagesGenerator = createReactComponentGenerator(
  ReactStyleVariation.CSSModules,
  {
    plugins: [headConfigPlugin],
    mappings: [ReactProjectMapping],
  }
);
```

We currently don't have a global place to configure the style variation in the strategy.
https://github.com/teleporthq/teleport-code-generators/blob/development/packages/teleport-types/src/generators.ts#L134

Let's introduce a new field in strategy something like this

```js
const generator = createProjectGenerator({
  components: {
    generator: reactComponentGenerator,
    path: ["src", "components"],
  },
  pages: {
    generator: reactPagesGenerator,
    path: ["src", "views"],
  },
  projectStyleSheet: {
    generator: styleSheetGenerator,
    fileName: "style",
    path: ["src"],
  },
  router: {
    generator: routingComponentGenerator,
    path: ["src"],
    fileName: "index",
  },
  entry: {
    generator: htmlFileGenerator,
    path: ["public"],
    fileName: "index",
  },
  static: {
    prefix: "",
    path: ["public"],
  },
});
```

to this

```js
const generator = createProjectGenerator({
  variation: ReactStyleVariation.CSSModules,
  components: {
    generator: reactComponentGenerator,
    path: ["src", "components"],
    plugins: [importStatementsPlugins],
    processors: [prettierJS],
    mappaings: [],
  },
  pages: {
    generator: reactPagesGenerator,
    path: ["src", "views"],
    plugins: [],
    processors: [],
  },
  projectStyleSheet: {
    generator: styleSheetGenerator,
    fileName: "style",
    path: ["src"],
    plugins: [],
    processors: [],
  },
  router: {
    generator: routingComponentGenerator,
    path: ["src"],
    plugins: [reactAppRoutingPlugin],
    processors: [],
    fileName: "index",
  },
  entry: {
    generator: htmlFileGenerator,
    path: ["public"],
    fileName: "index",
  },
  static: {
    prefix: "",
    path: ["public"],
  },
});
```

So, the appending plugins and post-processors and style variation configurations happens in project-generators.

JayaKrishnaNamburu:
### Question 1

Thinking a little more, we shouldn't allow template to be changed from a project plugin. In my view the main use case for project plugins is to quickly adding some extra component plugins or remove some. Or change some files that are generated along. 

In this case, if we build plugins on a default generator and some plugin changes the template in middle. Then the whole pipeline gives us a broken project. So, if we want to change the template, it can be done by building a new generator.

JayaKrishnaNamburu:
### Question 2

Should we add a new option in component-generators, to emit Code or just AST at the end of pipeline.

When we think of component-generators alone, having code at the end makes sense. But when we think of project-generators and the plugin eco-system around it. Sometimes, we need to change a couple of files or edit them or add them. 

Right now, because of

**AST to Code** and **Prettier** happening at component level. We are using MagicString to quickly add or edit a file as here https://github.com/teleporthq/teleport-code-generators/pull/514/files#diff-9fcbca22388001116d00da990af640ebed9d89ba33d07f1324fabb6b733830cdR32
and then running prettier on top of it.

The problem in long run is, Magic String works using indexes as we can see from the existing plugins. So every-time, even a little of context changes inside the component. These indexes need to be changed to support them. Which is not the best solution.

So my emitting AST from component generators during project-generators pipeline. We can just change the AST structure and the post-proceessors run at the end. Since we will be changing AST, we don't need to worry about indexes.

Utwo (approved):

Utwo:
Awesome work! :100:

JayaKrishnaNamburu:
Most of the plugins will be opinionated towards a framework. So, basically we can't use a `pluginNextCssModules` to a `gatsby-project-generator`. So, we need a `strategy-id` through which we can detect and throw error in plugins. If the `strategy-id` is not the plugin looking for. So, we can eliminate the errors, that we run into by using a plugin of one framework with another one. 

Or is it really mandatory for the plugins to do this check before doing operation ?

cc @Utwo

PR #515, opened 2020-12-01
chore(package.json): Adding stream flag for build script

JayaKrishnaNamburu:


JayaKrishnaNamburu:
@Utwo looks like the `stream` flag with lerna is cutting down the build time by 2-3mins at minimum --> https://github.com/teleporthq/teleport-code-generators/pull/514

Can we use this till we upgrade to `berry` and use https://yarn.build ?

Utwo (approved):

PR #528, opened 2021-01-05
fix(validator): Throwing error if the UIDL has 'null' or 'undefined'

### LINKED ISSUE #520: Clean UIDL before validating
We just take the JSON directly and apply the validators on top of it. 
https://github.com/teleporthq/teleport-code-generators/blob/development/packages/teleport-uidl-validator/src/validator/index.ts#L20

When we are integrating in external systems, there are high changes of so un-desired values in the JSON that we get. Like `undefined` `null` etc. Which cause the validators to break. Should we clean the UIDL on our side before doing the validation step is out for discussion.

Because, every system that is trying to integrate should clean the JSON themselves and then do the generation part. And if they do miss, at times they might get undesirable results and the error message is little too big to convey it.

```js
const parsedJSON = JSON.parse(JSON.stringify(uidl), (key, value) => {
  if (value !== null) {
    return value;
  }
});
```

JayaKrishnaNamburu:
fixes #520

### DIFF
--- a/packages/teleport-uidl-validator/__tests__/validator/component-uidl-with-null-undefined.json
+++ b/packages/teleport-uidl-validator/__tests__/validator/component-uidl-with-null-undefined.json
@@ -1,0 +1,24 @@
+{
+    "name": "Sample Component",
+    "node": {
+      "type": "element",
+      "content": {
+        "elementType": "div",
+        "name": "div",
+        "children": [
+          {
+            "type": "static",
+            "content": "Hello World"
+          }
+        ],
+        "attrs": {},
+        "style": {
+          "width": {
+            "type": "static",
+            "content": "100px"
+          },
+          "height": null
+        }
+      }
+    }
+  }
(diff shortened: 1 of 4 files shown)


JayaKrishnaNamburu:
`stringify` and `parse` will solve the issue. But with the current implementation. 

```json
"style": {
   "width": null
},
```

will  get converted into

```json
"style" : {}
```

But, any complex use cases like

```json
"style": {
     "type":  "static",
     "content": null
 },
```

will be converted to 

```json
"style": {
     "type":  "static"
 },
```

the validator throws error since it is invalid. So, one way is to parse through the parent and remove the ``type`` too. So, instead of parsing and trying to solve the error by ourself with more additational checks. We just through a `error` saying the value can't be ``null``. And let the users and integrations solve the error.

cc @Utwo

Utwo:
Yes, I agree on throwing an error when the validator find a `null` value.

Utwo (approved):

Utwo:
:100:

Issue #557, opened 2021-04-20
Projects render un-expectedly when they are exported, because of the custom tag wrapper around them


JayaKrishnaNamburu:
### Issue

When we use `components` in angular. We are fairly similar with the `selector` syntax like `app-component` or `app-home` etc. Which are similar to how `custom-elements` are defined. So, browsers don't have a clue on how to render these. And so defines them as `inline` elements. 

If an element is `inline` the width and height doesn't effect. And so, if the element is styled internally with a width and height the changes still doesn't have any effect. Let's take a look at an example from Angular

Codesandbox link --> https://codesandbox.io/s/bjkpo
<img width="1440" alt="Screenshot 2021-04-20 at 8 08 56 PM" src="https://user-images.githubusercontent.com/11075561/115415252-60158200-a214-11eb-9921-951a09ac7678.png">

The component root has a width of `100%` explicitly set by the user, but this doesn't reflect because. The encapsulated `app-component` behaves as a `inline` element.

CSS for the root element inside component

```css
.container {
  width: 100%;
  height: 400px;
  display: flex;
  position: relative;
  align-items: center;
  flex-direction: column;
  justify-content: center;
  background-color: var(--dl-color-success-300);
}

```

Let's check the same example using **React**
Codesandbox link --> https://codesandbox.io/s/7x28c
<img width="1437" alt="Screenshot 2021-04-20 at 8 08 48 PM" src="https://user-images.githubusercontent.com/11075561/115415747-c9959080-a214-11eb-8d80-9af546afa007.png">

## Solution

- The solution is always add `contents` to the `:host` of the component. So, the component wrapper doesn't effect the way in which things are rendered. Like here 
https://codesandbox.io/s/angular-bug-forked-4lcpk?file=/src/app/components/component/component.component.css
<img width="1112" alt="Screenshot 2021-04-20 at 8 27 16 PM" src="https://user-images.githubusercontent.com/11075561/115418089-d1563480-a216-11eb-820a-8c2d49446c46.png">

- The other way is to add `width` to the the `:host` when any width is explicitly mentioned on the root of the element.
```css
:host {
  display: block;
  width: 100%
}
```

My opinion is `contents` works the best. Since, with the second solution we might sometimes come across un-expected behaviours too.

JayaKrishnaNamburu:
Additional links 

https://nitayneeman.com/posts/solving-a-styling-issue-for-rendered-components-in-angular/
[...]

PR #558, opened 2021-04-21
fix(plugin-css): add contents by default for shadow DOM and custom elments based frameworks

JayaKrishnaNamburu:
#557


Utwo (approved):
:100:

Issue #277, opened 2019-06-14
Extend validation to events


anamariaoros:
Validation should be extended to event defined in UIDL.
Currently it throws this error:
~~~
TypeError: Cannot read property 'type' of undefined
    at createStateChangeStatement (utils.js:251)
    at utils.js:47
    at Array.forEach (<anonymous>)
~~~

To reproduce:
- use https://repl.teleporthq.io/
- enter the uidl below:
~~~
{
  "name": "Simple Component",
  "node": {
    "type": "element",
    "content": {
      "elementType": "container",
      "children": [{
        "type": "element",
        "content": {
          "elementType": "button",
          "name": "openmodal",
          "children": [{
                  "type": "static",
                  "content": "Show Popup"
                }],
          "events": {
            "click": [{
            "type": "stateChange",
            "modifies": "isOpen",
            "newState": true
          }]
          },
          "attrs": { 
            "data-static-attr": {
              "type": "static",
              "content": "test"
            }
          },
          "attributes":{ 
            "htmlFor": {
             "type": "static",
              "content": "test"
            }
          }
        }
      }]
    }
  }
}
~~~
- check error in console

chavda-bhavik:
Hii, I want to contribute to this. 

I get to know that this error occurred when using react, so I look into `packages/teleport-component-generator-react`, but couldn't able to find anything.

Can you say where to look for this issue?

JayaKrishnaNamburu:
Hello @chavda-bhavik happy to have you here, UIDL is validated before passing for `code-generaion`. So, fix should happen in `teleporthq/teleport-udl-validator`. I will try to explain the issue here. 

- Here is an example which uses events - https://repl.teleporthq.io/?uidlLink=fa289c0f-f544-4fcf-ae48-b1e0f953596c&flavor=React&style=CSS-Modules
- So, for an event handler, right now we support only a state change. So, we need a valid `stateDefiniton` in the event handler.
-  modified the above UDL from the PR here - https://repl.teleporthq.io/?uidlLink=495dcf70-320f-472c-a20e-7031d1bd1722&flavor=React&style=CSS-Modules
- The bug here is, validator should throw error, when `isOpen` stateDefinition is not definied.
- You can check this https://github.com/teleporthq/teleport-code-generators/blob/development/packages/teleport-uidl-validator/src/validator/utils.ts#L160 as an example for this.

chavda-bhavik:
[...]

PR #560, opened 2021-04-21
(refactor): Extending validations for events in UIDL

chavda-bhavik:
I've performed checks by doing these edits to the following files, to fix this [issue](https://github.com/teleporthq/teleport-code-generators/issues/277)

**/packages/teleport-test/src/standalone.ts**
```
...
import componentJSON from '../../../examples/uidl-samples/component.json'
import Validator from '../../teleport-uidl-validator/src/validator'
const componentUIDL = (componentJSON as unknown) as ComponentUIDL;
...
const run = async () => {
 try {
  ...
  let validator = new Validator();
  result = validator.validateComponentContent(componentUIDL);
  console.log(result);
 } catch(e) {
  ...
 }
}
```

/examples/uidl-samples/component.json
```
{
  "name": "Simple Component",
  "node": {
    "type": "element",
    "content": {
      "elementType": "container",
      "children": [{
        "type": "element",
        "content": {
          "elementType": "button",
          "name": "openmodal",
          "children": [{
                  "type": "static",
                  "content": "Show Popup"
                }],
          "events": {
            "click": [{
            "type": "stateChange",
            "modifies": "isOpen",
            "newState": true
          }]
          },
          "attrs": { 
            "data-static-attr": {
              "type": "static",
              "content": "test"
            }
          },
          "attributes":{ 
            "htmlFor": {
             "type": "static",
              "content": "test"
            }
          }
        }
      }]
    }
  }
}
```


packages/teleport-uidl-validator/src/validator/utils.ts
```diff
@@ -99,6 +99,17 @@ export const checkDynamicDefinitions = (input: Record<string, unknown>) => {
   const errors: string[] = []
 
   UIDLUtils.traverseNodes(input.node as UIDLNode, (node) => {
+    if (node.type === 'element' && node.content.events) {
+      Object.keys(node.content.events).forEach((eventKey) => {
+        node.content.events[eventKey].forEach((event) => {
+          if (!stateKeys.includes(event.modifies)) {
+            const errorMsg = `"${event.modifies}" is used in events, but not defined in state. Please add it in stateDefinitions`
```
JayaKrishnaNamburu: Please update the error message to 

```js
`"${event.modifies}" is used in events, but not defined. Please add it in stateDefinitions`
```

Because, we currently support event modifiers only for stateDefinitions.

JayaKrishnaNamburu: Can we also add an additional check if the `modifies` is defined in `stateDefinitions`.

```js
typeof event.newState === input.stateDefinitions[event.modifies]?.type
```

So, we are sure that the newState is in sync with the `stateDefinition` in UIDL.

JayaKrishnaNamburu (changes requested):
Thanks @chavda-bhavik, for your contribution. Just some minor improvements. And can we add `__tests__` too for the validation please 😄

packages/teleport-uidl-validator/__tests__/validator/component-uidl-with-state-undefined-mismatch.json
```diff
@@ -0,0 +1,49 @@
+{
```
JayaKrishnaNamburu: Please use properly formatted json. You can use prettier-vscode or you can use this tool to make it pretty https://jsonformatter.org/json-pretty-print

```json
{
  "name": "Simple Component",
  "stateDefinitions": {
    "isOpen": {
      "type": "boolean",
      "defaultValue": false
    }
  },
  "node": {
    "type": "element",
    "content": {
      "elementType": "container",
      "children": [
        {
          "type": "element",
          "content": {
            "elementType": "button",
            "name": "openmodal",
            "children": [
              {
                "type": "static",
                "content": "Show Popup"
              }
            ],
            "events": {
              "click": [
                {
                  "type": "stateChange",
                  "modifies": "isOpen",
                  "newState": "test"
                },
                {
                  "type": "stateChange",
                  "modifies": "isExplored",
                  "newState": true
                }
              ]
            },
            "attrs": {
              "data-static-attr": {
                "type": "static",
                "content": "test"
              }
            },
            "attributes": {
              "htmlFor": {
                "type": "static",
                "content": "test"
              }
            }
          }
        }
      ]
    }
  }
}
```

packages/teleport-uidl-validator/__tests__/validator/index.ts
```diff
@@ -1,5 +1,7 @@
 import { Validator } from '../../src'
 
+// @ts-ignore
+import componentUidlWithStateUndefinedMismatch from './component-uidl-with-state-undefined-mismatch.json'
```
JayaKrishnaNamburu: Please change the file name to `component-uidl-with-event-modifier-mismatch.json`

packages/teleport-uidl-validator/src/validator/utils.ts
```diff
@@ -99,6 +99,24 @@ export const checkDynamicDefinitions = (input: Record<string, unknown>) => {
   const errors: string[] = []
 
   UIDLUtils.traverseNodes(input.node as UIDLNode, (node) => {
+    if (node.type === 'element' && node.content.events) {
```
JayaKrishnaNamburu, line 102: Looks good 👍, but please don't use // @ts-ignore in the code. Tests it's fine, since we are importing `.json` files. And the checks and error message can be improved a little.

```suggestion
    if (node.type === 'element' && node.content.events) {
      Object.keys(node.content.events).forEach((eventKey) => {
        node.content.events[eventKey].forEach((event) => {
          if (!stateKeys.includes(event.modifies)) {
            const errorMsg = `"${event.modifies}" is used in events, but not defined`
            errors.push(errorMsg)
            return
          }
      })
    }
```

JayaKrishnaNamburu (changes requested):
Thanks for your time, just some file-name changes, formatting improvements and we are good to go 😄

packages/teleport-uidl-validator/src/validator/utils.ts
JayaKrishnaNamburu: Let's remove the type-check. Sorry for that. Here is the reason, we have two type of handlers defined in event. One is `toggle` type that modifies a state. And second is an event handler. So, for now, let's remove the type check

```js
{
  "click": [
    {
      "type": "propCall",
      "calls": "onClose"
    },
    {
      "type": "stateChange",
      "modifies": "fakeState",
      "newState": false
    }
  ]
}
```

As you can see, the handlers of two types. `stateChange` and `propCall`. The check for `modifies` should happen when it is `stateChange` and the `newState` is not equal to `$toggle`

If it is `$toggle` the state that modifies should be of `boolean`. Since, we can toggle only a boolean.
```js
{
  "click": [
    {
      "type": "stateChange",
      "modifies": "isExpanded",
      "newState": "$toggle"
    }
  ]
}
```

And one more last thing, can you try extending the `decoders` to check for events too. But that's optional 😄. I can take care if you want.  But here is where you can take a look
https://github.com/teleporthq/teleport-code-generators/blob/development/packages/teleport-uidl-validator/src/decoders/utils.ts#L287
You can define `propCall` and `stateChange` events and make a union out of it. 

The events area of the project is not documented so great. So, it will be a good start and great help with the PR. Thank you 😄

Feel free to ping anytime for any doubts, and you can run `yarn test:watch` or `yarn test` to cross-check the test-cases 😄

chavda-bhavik: I'll be happy to do that @JayaKrishnaNamburu, I think union events will be defined as
`
export const eventHandlerStatementDecoder: Decoder<UIDLEventHandlerStatement> = object({
  type: union(constant("stateChange"), constant('propCall')),
  calls: optional(string()),
  modifies: optional(string()),
  newState: optional(union(string(), number(), boolean())),

  args: optional(array(union(string(), number(), boolean()))),
})
`
right?

JayaKrishnaNamburu: This works, but not intuitive at first glance or working with it. It will be more clear if we approach this something like

in teleport-types/uidl.ts

```js
export type UIDLEventHandlerStatement = UIDLPropCallEvent | UIDLStateModifierEvent
```

```js
export interface UIDLPropCallEvent {
  type: 'propCall'
  calls?: string
  args?: Array<string | number | boolean>
}
export interface UIDLStateModifierEvent {
  type: 'stateChange'
  modifies?: string
  newState?: string | number | boolean
}

export type UIDLEventHandlerStatement = UIDLPropCallEvent | UIDLStateModifierEvent
```

And then extend on the `validators` and the same for the validations inside the `validators` in `validateComponentSchema` too.

chavda-bhavik:
summary of changes made:

- added types for propCall and stateChange events
- added stateDefinitions to tests of vue-angular component generators, they were failing because of undefined state
- applied type **UIDLPropCallEvent** and **UIDLStateModifierEvent** in files which were handling propCall and stateChange events
- removed state type checking from `teleport-uidl-validator/src/validator/utils.ts` and `teleport-uidl-validator/__tests__/validator/index.ts` 
- renamed test-event-component-uidl json file to `component-uidl-with-event-modifier-undefined.json`
- added union of propCall and stateChange event in `teleport-uidl-validator/src/decoders/utils.ts`

JayaKrishnaNamburu:
Awesome work @chavda-bhavik first glance everything looks solid. Will take a detailed look tomorrow. Once again, thanks for your time.

JayaKrishnaNamburu (approved):
Thanks @chavda-bhavik merging it, i just made some small changes that's it 😄

PR #561, opened 2021-04-23
feat(validator): Extending support for element-state validations

oanamun:


packages/teleport-types/src/uidl.ts
```diff
@@ -343,7 +343,7 @@ export interface UIDLStyleStateCondition {
   content: UIDLElementStyleStates
 }
 
-export type UIDLElementStyleStates = 'hover' | 'active' | 'focus' | 'disabled'
+export type UIDLElementStyleStates = 'hover' | 'active' | 'focus' | 'focus-within' | 'focus-visible' |  'disabled' | 'visited' | 'checked' | 'placeholder'
```
JayaKrishnaNamburu, line 346: Please add validations for these in the validators :)
You can check the validators here --> https://github.com/teleporthq/teleport-code-generators/blob/development/packages/teleport-uidl-validator/src/decoders/utils.ts#L344

JayaKrishnaNamburu (changes requested):

packages/teleport-uidl-validator/src/decoders/utils.ts
```diff
@@ -345,7 +345,12 @@ export const elementStateDecoder: Decoder<UIDLElementStyleStates> = union(
   constant('hover'),
```
JayaKrishnaNamburu, line 346:
```suggestion
export const elementStateDecoder: Decoder<UIDLElementStyleStates> = oneOf(
  constant('hover'),
```

and import `oneOf` instead of union. from `@mojotech/json-type-validation`

JayaKrishnaNamburu (changes requested):
Please change from `union` to use `oneOf` so it behaves as enum check.

JayaKrishnaNamburu (approved):
💯

PR #563, opened 2021-04-28
fix(generator): enable jsec encoding to support special characters in AST to code transformation

JayaKrishnaNamburu:
Please visit the following URL --> https://repl.teleporthq.io/project/cc023c27-ba39-449a-8d97-013e616256c3

You can see, when a component is passed with props that contains non-english characters. With `jsec` in minimal. It will stop encoding them in babel.

## Result
Screenshot 2021-04-28 at 4 59 07 PM


Utwo (approved):
:rocket:

PR #564, opened 2021-05-02
Improve UIDL validation for global assets

### LINKED ISSUE #457: Improve UIDL validation for global assets
For handling style tags, we need a string content or a path. But not both at the same time

```json
      {
        "type": "style",
        "content": " a {  text-decoration: underline; } button {  background-color: #fff; border-radius: 4px; border-color: #0089be; border-width: 1px; padding: 5px 10px; text-align: center; } input[type=text] {  border-color: #0089be; border-radius: 2px; border-width: 1px; } input[type=number] {  border-color: #0089be; border-radius: 2px; border-width: 1px; } textarea {  border-color: #0089be; border-radius: 2px; border-width: 1px; } svg {  fill: #2b2b2b; border-color: #2b2b2b; }"
      },
      {
        "type": "style",
        "path": "https://unpkg.com/pattern.css"
      }
```

Right now, we don't have a proper check. You can pass both content and path for the UIDL and it still works. But generates two tags. But ideally we should not allow this behaviour. Style tag either contains an external link or some self content.

- Supporting additational attributes or link tags like, `cross-origin` `integrity` etc will be a added advantage

chavda-bhavik:
For issue #457 I've added types for each asset and made changes based on that. But I'm confused about where to write the logic for "For handling style tags, we need a string content or a path. But not both at the same time"?

### DIFF
--- a/packages/teleport-plugin-common/src/builders/ast-builders.ts
+++ b/packages/teleport-plugin-common/src/builders/ast-builders.ts
@@ -188,7 +188,10 @@ export const appendAssetsAST = (
   bodyNode: types.JSXElement
 ) => {
   assets.forEach((asset) => {
-    const assetPath = UIDLUtils.prefixAssetsPath(options.assetsPrefix, asset.path)
+    let assetPath
+    if ('path' in asset) {
+      assetPath = UIDLUtils.prefixAssetsPath(options.assetsPrefix, asset.path)
+    }
 
     // link canonical for SEO
     if (asset.type === 'canonical' && assetPath) {
@@ -207,7 +210,7 @@ export const appendAssetsAST = (
     }
 
     // inline style
-    if (asset.type === 'style' && asset.content) {
+    if (asset.type === 'style' && 'content' in asset) {
       const styleTag = createJSXTag('style')
       addAttributeToJSXTag(styleTag, 'dangerouslySetInnerHTML', { __html: asset.content })
       addChildJSXTag(headNode, styleTag)
@@ -217,6 +220,7 @@ export const appendAssetsAST = (
     if (asset.type === 'script') {
       const scriptTag = createJSXTag('script')
       addAttributeToJSXTag(scriptTag, 'type', 'text/javascript')
+
       if (assetPath) {
         addAttributeToJSXTag(scriptTag, 'src', assetPath)
         if (asset.options && asset.options.defer) {
@@ -225,7 +229,7 @@ export const appendAssetsAST = (
         if (asset.options && asset.options.async) {
           addAttributeToJSXTag(scriptTag, 'async', true)
         }
-      } else if (asset.content) {
+      } else if ('content' in asset) {
         addAttributeToJSXTag(scriptTag, 'dangerouslySetInnerHTML', {
           __html: asset.content,
         })
--- a/packages/teleport-plugin-common/src/builders/ast-builders.ts
+++ b/packages/teleport-plugin-common/src/builders/ast-builders.ts
@@ -188,10 +188,7 @@ export const appendAssetsAST = (
   bodyNode: types.JSXElement
 ) => {
   assets.forEach((asset) => {
-    let assetPath
-    if ('path' in asset) {
-      assetPath = UIDLUtils.prefixAssetsPath(options.assetsPrefix, asset.path)
-    }
+    const assetPath = UIDLUtils.prefixAssetsPath(options.assetsPrefix, asset.path)
 
     // link canonical for SEO
     if (asset.type === 'canonical' && assetPath) {
@@ -210,7 +207,7 @@ export const appendAssetsAST = (
     }
 
     // inline style
-    if (asset.type === 'style' && 'content' in asset) {
+    if (asset.type === 'style' && asset.content) {
       const styleTag = createJSXTag('style')
       addAttributeToJSXTag(styleTag, 'dangerouslySetInnerHTML', { __html: asset.content })
       addChildJSXTag(headNode, styleTag)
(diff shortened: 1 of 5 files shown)


packages/teleport-plugin-common/src/builders/ast-builders.ts
```diff
@@ -188,7 +188,10 @@ export const appendAssetsAST = (
   bodyNode: types.JSXElement
 ) => {
   assets.forEach((asset) => {
-    const assetPath = UIDLUtils.prefixAssetsPath(options.assetsPrefix, asset.path)
```
JayaKrishnaNamburu, line 191: Please use [optional-chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) `asset?.path`. We follow the similar everywhere

packages/teleport-project-generator/src/file-handlers.ts
```diff
@@ -240,26 +243,31 @@ const createHTMLEntryFileChunks = (uidl: ProjectUIDL, options: EntryFileOptions)
...
 
     // script (external or inline)
     if (asset.type === 'script') {
-      const scriptInBody = (asset.options && asset.options.target === 'body') || false
+      let scriptInBody = false
+      if ('options' in asset) {
+        if ('target' in asset.options && asset.options.target === 'body') {
+          scriptInBody = true
+        }
+      }
```
JayaKrishnaNamburu:
```suggestion
      const scriptInBody = asset?.options && asset?.options?.target === 'body' ? true  : false
```

packages/teleport-project-generator/src/file-handlers.ts
```diff
@@ -240,26 +243,31 @@ const createHTMLEntryFileChunks = (uidl: ProjectUIDL, options: EntryFileOptions)
...
-        if (asset.options && asset.options.defer) {
+        if ('options' in asset && asset.options.defer) {
           HASTUtils.addBooleanAttributeToNode(scriptTag, 'defer')
         }
-        if (asset.options && asset.options.async) {
+        if ('options' in asset && asset.options.async) {
           HASTUtils.addBooleanAttributeToNode(scriptTag, 'async')
         }
-      } else if (asset.content) {
+      } else if ('content' in asset) {
```
JayaKrishnaNamburu, line 266: same syntax as above

JayaKrishnaNamburu (changes requested):
Just some changes

packages/teleport-project-generator/src/file-handlers.ts
chavda-bhavik: @JayaKrishnaNamburu  `asset?.content` throws typescript error that `Property 'content' does not exist on type 'UIDLScriptAsset'.  Property 'content' does not exist on type 'UIDLScriptExternalAsset'`. This happens for others too.

JayaKrishnaNamburu:
> For issue #457 I've added types for each asset and made changes based on that. But I'm confused about where to write the logic for "For handling style tags, we need a string content or a path. But not both at the same time"?

With separated validations like this 
```js
export interface UIDLStyleExternalAsset {
  type: 'style'
  path: string
}
export interface UIDLStyleInlineAsset {
  type: 'style'
  content: string
}
```

If we get a input of 
```js
{
  "type": "style",
  "path": "some-path",
  "content": "some-content"
}
```

The type checks will take care of the inputs. Because, a style tag can't have both `content` and `external-link` with at the same time. The generated-code is still good. Since, the first line finds it as external link and generates a link tag and the second generates a style tag.

packages/teleport-project-generator/src/file-handlers.ts
JayaKrishnaNamburu: It should be something like this 

```js
if (asset.type === 'script') {
      const scriptInBody = (asset.options && asset.options.target === 'body') || false
      const scriptTag = HASTBuilders.createHTMLNode('script')
      HASTUtils.addAttributeToNode(scriptTag, 'type', 'text/javascript')

      if (assetPath && asset.path) {
        HASTUtils.addAttributeToNode(scriptTag, 'src', assetPath)
      }

      if (assetPath && asset.content) {
        HASTUtils.addTextNode(scriptTag, asset.content)
      }

      if (asset.options && asset.options.defer) {
        HASTUtils.addBooleanAttributeToNode(scriptTag, 'defer')
      }
      if (asset.options && asset.options.async) {
        HASTUtils.addBooleanAttributeToNode(scriptTag, 'async')
      }

      if (scriptInBody) {
        HASTUtils.addChildNode(bodyNode, scriptTag)
      } else {
        HASTUtils.addChildNode(headNode, scriptTag)
      }
    }
```

And by the way, `options` should be present for `UIDLScriptInlineAsset`. Because, any script tag can have options `async` | `defer` etc. But as the same as styles, it can't have both `src` of remote and script content at the same time. It's just a `syntatical` difference but mostly they have the same options and everything 😄

packages/teleport-types/src/uidl.ts
```diff
@@ -19,18 +19,54 @@ export interface UIDLGlobalProjectValues {
...
   options?: {
     async?: boolean
     defer?: boolean
     target?: string
+  }
+}
+export interface UIDLScriptInlineAsset {
+  type: 'script'
+  content: string
+}
```
JayaKrishnaNamburu, line 34:
```suggestion
export interface UIDLScriptExternalAsset {
  type: 'script'
  path: string
  options?: {
    async?: boolean
    defer?: boolean
    target?: string
  }
}
export interface UIDLScriptInlineAsset {
  type: 'script'
  content: string,
  options?: {
    async?: boolean
    defer?: boolean
    target?: string
  }
}
```

It's just that we can't have both content and path at the same time 😄

JayaKrishnaNamburu:
@all-contributors please add @chavda-bhavik  for code

JayaKrishnaNamburu (approved):
@chavda-bhavik i made some changes, merging it. Thanks for your contributions 🚀

PR #570, opened 2021-05-09
improve validations to return empty element name as valid value

### LINKED ISSUE #567: Improve validations for checking `name` on `elementNode`
https://github.com/teleporthq/teleport-code-generators/blob/development/packages/teleport-uidl-validator/src/decoders/utils.ts#L472

Right now, here we check if the UIDL's name is valid. But when we pass a `empty string` the validation breaks. `name` field is not mandatory, but don't throw an error. Instead just ignore the name

https://repl.teleporthq.io/?uidlLink=02853c0e-b6dd-45fa-b9e5-74bd6f3d7890&flavor=React&style=CSS-Modules

chavda-bhavik:
For #567 I've modified `isValidElementName()` to return empty element name as valid value and updated regex now returns true for empty string.


packages/teleport-uidl-validator/src/decoders/custom-combinators.ts
```diff
@@ -43,8 +43,8 @@ export class CustomCombinators<A> {
 
   static isValidElementName(): CustomCombinators<string> {
     return new CustomCombinators<string>((json: string) => {
-      const fileNameRegex = new RegExp('^[a-zA-Z]+[a-zA-Z0-9-_]*$')
-      if (json && typeof json === 'string' && fileNameRegex.test(json)) {
+      const fileNameRegex = new RegExp('^$|^[a-zA-Z]+[a-zA-Z0-9-_]*$')
+      if (typeof json === 'string' && fileNameRegex.test(json)) {
```
JayaKrishnaNamburu: Well, `empty-strings` are not valid element names. So, it will be great if we can do thr string length check and throw error if it is empty strings.

packages/teleport-uidl-validator/src/decoders/custom-combinators.ts
```diff
@@ -43,8 +43,8 @@ export class CustomCombinators<A> {
...
         return Result.ok(json)
       }
       return Result.err({ message: `Invalid link attribute, received ${json}` })
```
JayaKrishnaNamburu: Here, can you please update the message to `Invalid name`, instead of `link attribute`

JayaKrishnaNamburu (changes requested):

packages/teleport-uidl-validator/src/decoders/custom-combinators.ts
chavda-bhavik: @JayaKrishnaNamburu, means we need to throw an error for empty element name? was thinking that the empty element name is valid, and made change accordingly.

JayaKrishnaNamburu: Yes, actually `name` field is optional. But if the field exists then it should be a valid name. So, `name` can't be a empty string.

chavda-bhavik: I think that thing was already happening before, throwing an error if element `name` has an empty string, and it's optional.

JayaKrishnaNamburu: Yes, the issue was to improve error message 😃

chavda-bhavik:
@JayaKrishnaNamburu validation throws error message as below, is this issue?
`      
 Error: UIDL Format Validation Error. Please check the following:
        - Path input.node.content.name: [object Object].
        is a DecoderError
`

JayaKrishnaNamburu:
@chavda-bhavik this is a error message that is already happening. The aim was to make it more easy to pin-point the type of error for the user 😄. Can you try by throwing the error directly with `throw new Error` instead of `Result.Error`. Because, anyways `Resut.Error` breaks the pipeline anyways. What you say ?

chavda-bhavik:
Yes, that would be great @JayaKrishnaNamburu. I will update that way.

JayaKrishnaNamburu:
Awesome, to explore this path will be a great help

chavda-bhavik:
@JayaKrishnaNamburu by throw new Error(''), we can see the message as below, can we go with this? asking because `undefined` is coming for the path and for `DecoderError` or `ValidationError`.
`

UIDL Format Validation Error. Please check the following: 
     - Path undefined: Error: Name attribute cannot be empty. 
     is a undefined 
`

JayaKrishnaNamburu:
> @JayaKrishnaNamburu by throw new Error(''), we can see the message as below, can we go with this? asking because `undefined` is coming for the path and for `DecoderError` or `ValidationError`.
> `
> 
> UIDL Format Validation Error. Please check the following:
> - Path undefined: Error: Name attribute cannot be empty.
> is a undefined
> `

Sounds good, can we repeat the same for isValidComponentName and isValidProjectName too please :)

JayaKrishnaNamburu (dismissed):

JayaKrishnaNamburu:
Thanks @chavda-bhavik for the PR 😄 💯

JayaKrishnaNamburu (approved):

PR #572, opened 2021-05-14
feat/cli

JayaKrishnaNamburu:


packages/teleport-mapper/src/studio/constatnts.ts
```diff
@@ -0,0 +1,117 @@
+export const DEFAULT_THEME_CONSTANTS = {
```
Utwo, line 1: Typo in filename

Utwo: Typo in filename

Utwo (changes requested):

packages/teleport-cli/Readme.md
```diff
@@ -0,0 +1,63 @@
...
+
+teleport clone --link https://teleport-gui-git-development-teleport-team.vercel.app/projects/new-project-zan7
+```
+
+### Clone a Component
+
+```shell
+teleport clone --link <link-to-component> --path <sys-path>
+
+teleport clone --link https://repl.teleporthq.io/\?uidlLink\=35c82f1c-d272-4f8b-8a3f-5fa44702a1e6\&flavor\=React\&style\=CSS-Modules
```
Utwo, line 18: Do we need to escape characters in the URL?

packages/teleport-cli/src/constants.ts
```diff
@@ -0,0 +1,23 @@
...
+export const CONIFG_FILE_EXTENSION = '.json'
+export const DEFAULT_CONFIG_FILE_NAME = `${CONFIG_FILE_NAME}${CONIFG_FILE_EXTENSION}`
+
+export const DEFALT_CONFIG_TEMPLATE: DefaultConfigTemplate = {
+  project: {},
+  components: {},
+}
+
+export const BASE_URL = 'https://us-central1-croapp-dev.cloudfunctions.net/repl-api/'
+export const STUDIO_URL = 'https://playground-api-production-v5.services.teleporthq.io/project/'
```
Utwo: Shouldn't this URL be the frontend one?

Utwo: If not, it's a little tricky, because we might upgrade the backend in the future, let's say from v5 to v6.

Utwo: So in that case, an old version of the CLI will not work anymore

packages/teleport-cli/src/services/index.ts
```diff
@@ -0,0 +1,123 @@
...
+  UUDID_REGEX,
+  CONFIG_FILE_NAME,
+  DEFALT_CONFIG_TEMPLATE,
+  DefaultConfigTemplate,
+  DEFAULT_CONFIG_FILE_NAME,
+} from '../constants'
+import { getPackageJSON } from '../utils'
+
+export const fetchUIDLFromREPL = async (url: string): Promise<Record<string, unknown>> => {
+  const id = url.match(UUDID_REGEX)[0]
```
Utwo: I think here is better to use this https://www.npmjs.com/package/uuid#uuidvalidatestr

Utwo (changes requested):

Utwo:
And please a minimal Readme in teleport-mapper

packages/teleport-cli/Readme.md
JayaKrishnaNamburu: Not needed, this is just i copied from my terminal as is. You can just copy from `url` bar

PR #575, opened 2021-05-17
feat(next-image): a JSX plugin to use next/image for loading images efficiently

### LINKED ISSUE #574: Use `Image` from next/image for when a specific node inside uidl respects the requirements.
NextJS recently released NextJS 10 and a new Image-component to handle images. But it has some specific requirements. So, any node that matches the specific requirements. Please change it’s tag name to `Image` Instead of `img`.

https://nextjs.org/docs/api-reference/next/image

This can be achieved by passing a new plugin for `teleport-project-generator-next`. Since we don’t need to add framework specific business logics inside react-base component. A inspiration for this plugin could be `InlineStyles` Which we can check here https://github.com/teleporthq/teleport-code-generators/blob/development/packages/teleport-plugin-jsx-inline-styles/src/index.ts

But before this switch make sure the next templates are upgraded to Next-10. At the moment we are using Next-9.3 for generating projects.

JayaKrishnaNamburu:
fixes #574 

- Adding the `next/Image` if the image has `width` and `height` alone and of same css unit.
- Upgrading to Next@10 for the project-generation

### DIFF
--- a/packages/teleport-plugin-jsx-next-image/__tests__/index.ts
+++ b/packages/teleport-plugin-jsx-next-image/__tests__/index.ts
@@ -1,0 +1,74 @@
+import { component, elementNode, staticNode } from '@teleporthq/teleport-uidl-builders'
+import { createNextImagePlugin } from '../src'
+import {
+  ChunkDefinition,
+  ChunkType,
+  FileType,
+  ComponentStructure,
+} from '@teleporthq/teleport-types'
+import * as types from '@babel/types'
+
+describe('plugin-jsx-next-image', () => {
+  const plugin = createNextImagePlugin()
+
+  const componentChunk: ChunkDefinition = {
+    name: 'jsx-component',
+    meta: {
+      nodesLookup: {
+        container: {
+          openingElement: {
+            name: {
+              name: 'img',
+            },
+            attributes: [],
+          },
+          closingElement: {
+            name: {
+              name: 'img',
+            },
+          },
+        },
+      },
+    },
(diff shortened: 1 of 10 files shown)


JayaKrishnaNamburu:
@Utwo i made the changes to add `width` and `height` but here is the problem. Any styles that are directly applied on the image are being lost. Here is a codesandbox example, the first `img` tags relies on a class `default` and it uses `cover`. If we try to change it using `Image` we basically lost the other attributes on the image like `object-fit` etc. Looks like it's a problem that is being faced in general. 
https://codesandbox.io/s/exciting-bas-sn5cj?file=/pages/index.js

https://github.com/vercel/next.js/discussions/18312

Utwo:
Manage to get it working like this but the process is quite complicated. https://codesandbox.io/s/flamboyant-yonath-76ko0?file=/pages/index.js

From the github issue:

> If you want to style the img element itself, classes are working fine, the problem is that img has internal inline styles that will have a higher CSS priority than yours. This is the CSS that overwrites:

```
visibility: inherit;
position: absolute;
top: 0px;
left: 0px;
bottom: 0px;
right: 0px;
box-sizing: border-box;
padding: 0px;
border: none;
margin: auto;
display: block;
width: 0px;
height: 0px;
min-width: 100%;
max-width: 100%;
min-height: 100%;
max-height: 100%;
```

Maybe we can leave this as a draft and wait for an update from Vercel team.

Utwo (approved):
:rocket:

PR #577, opened 2021-05-22
(feat/esbuild) Custom scripts for build and watch using `esbuild`. And generating `types` using `tsc`

### LINKED ISSUE #466: Add exports field to all the packages
We do export the `esm` bundles using `module` field in the `package.json`. But it is not supported in `node` and `jspm`  and `skypack` relies on that.  Further details about how this `exports` field is helping us can be read here --> https://docs.skypack.dev/lookup-urls/lookup-urls#lookup-a-package-export

How skypack even takes advantage of this field and exports the necessary functions alone. Please add the `exports` field to all our packages to take advantage of such features.

### LINKED ISSUE #543: Audit dependencies for all packages.
We are defining `@teleporthq/teleport-shared` in dependencies for `@teleporthq/teleport-project-generator-react` but we are not using it. Please clear the un-used dependencies from the packages.

https://github.com/teleporthq/teleport-code-generators/blob/development/packages/teleport-project-generator-react/package.json#L38

JayaKrishnaNamburu:
fixes #543 as well

From now, 
- we will be transpiling only to `es6` and not down till `es5`. Main reason is `es5` is not supported by `esbuild` and second end users can still transpile down to `es5` if needed. 
- And the packages by served using `commonjs-es6` and `esm-es6`. If needed we can add `esnext` in future.
-  Audit for several missing dependencies


JayaKrishnaNamburu:
15sec for building both `esm` and `cjs` 🤯

scripts/build.mjs
```diff
@@ -0,0 +1,50 @@
...
+
+  if (!isEntryExists) {
+    throw new Error(`Entry file missing from ${packageName}`)
+  }
+
+  await build({
+    entryPoints: [`${path}/src/index.ts`],
+    outdir: `${path}/dist/cjs`,
+    format: 'cjs',
+    plugins: [new NodeModulesPolyfillPlugin()],
```
JayaKrishnaNamburu: - [x] Don't add polyfills in these builds, the end users djoukd be able to sdd them depending on the tsrger platform. Loke repl, where we polyfill path and fs explicitly.
- [x] add build target as es6, anyways es5 is not supported by esbuild. But this helps in transpiling any higher js syntaxes to es6.
- [x] Add export map in all packages

JayaKrishnaNamburu:
Finally, its down from from 10-13mins to less than 5min mark 😄 🚀 
Most recent reference - https://github.com/teleporthq/teleport-code-generators/pull/572/checks?check_run_id=2640078182

JayaKrishnaNamburu:
@Utwo  and @chavda-bhavik it will be great if you can test this in local 😄 . Node~14 is required to work with the repo. So, please update to node-14 and higher

scripts/build.mjs
```diff
@@ -0,0 +1,69 @@
...
+    return
+  }
+
+  if (!isEntryExists || !packageJSON) {
+    throw new Error(`Entry file missing from ${packageName}`)
+  }
+
+  const external = [...Object.keys(JSON.parse(packageJSON)?.dependencies || {})]
+  external.push('path')
+  external.push('fs')
```
JayaKrishnaNamburu, line 39: So, i am specifying `path` and `fs` also as external imports. Even though our main target is browsers but we can run in node too. But if we `polyfill` because of that. It's more of duplication in node environments. 
Anyways we are not polyfilling them before too.

Polyfill happens in which ever environment it is being used. like in repl we do explicitly
https://github.com/teleporthq/teleport-repl/blob/development/next.config.js#L5

Utwo (approved):

Utwo (dismissed):
Just tested with node 15 and everything works extremely fast! Really impressive work! :rocket: :zap:

Utwo (approved):

PR #584, opened 2021-06-22
fix(resolver): Resolve attrs on buttons and elements with links

### LINKED ISSUE #582: Attributes set on the node of `elementType` text with `abilities` are lost in the resolving phase.
When a text node has abilities, the node that is wrapped should retain the `attrs`. Here is the reproducing url. We can see the id `ta-top-menu-sticky-studio` misses from the generated code.

https://repl.teleporthq.io/?uidlLink=6ba61828-9e3a-4c89-84f7-6dfa2519b3c0&flavor=React&style=CSS-Modules

JayaKrishnaNamburu:
fixes #582

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
   <img src="https://github.com/teleporthq/teleport-code-generators/blob/development/Default.png" width="250"/>
 </p>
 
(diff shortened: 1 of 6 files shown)


JayaKrishnaNamburu:
@all-contributors please add @mihaisampaleanu for bug

PR #586, opened 2021-06-23
CJS and ESM for Node and Browser environments

JayaKrishnaNamburu:
- [x]  Use Interop pattern for importing and using `cjs` modules inside `esm`.  Since, we are exposing `module` under `exports too. Now build tools pick only `esm` and these causes interop issues for us.
- [x]  Most of the time, browser is the target for the packages. So, switch to adding polyfills by default instead of relying on the build tools.

#583


JayaKrishnaNamburu:
`jss` imports have been a little tricky to support the same imports in `node` and `browsers`. I made a PR here https://github.com/cssinjs/jss/pull/1528

JayaKrishnaNamburu:
- [x] Don't switch to `type: module` yet. Instead export both `cjs ` and `esm` and have `.mjs` for esm extension.

JayaKrishnaNamburu:
There are a multiple problems in switching to new style of development with esbuild mainly for monorepos.

- In esbuild, with every supported syntax, we can transpile all the way down only to `es6` many `es5` features are experimental and many are not supported yet.
- With `webpack5` and `es6` in node_modules. There is a bug within terser, which is not confirmed by the team yet. So it breaks in the optimisation pipeline of webpack. Which you can reproduce using https://stackblitz.com/edit/nextjs-uaz4wy?file=next.config.js. Webpack 5 along with `next` tends to create these broken bundles. Which works well when used with webpack4 https://stackblitz.com/edit/nextjs11-webpack4

- When we switch a package to use `type :module` all the files inside are considered as `esm` modules. And in `esm` modules we always need to specify the extension along with the imports. The problem is `typescript` can transpile but can't bunch the files to form a build. It can do that only with `systemjs` and `iife`.
   - So, the other option to do that, is use `rollup` or `esbuild` to bundle all the files. Remember, `esbuild` can do till `es6` and so it is out of option now. Second, using `rollup`, and rollup along side with `rollup-plugin-typescript2` can solve this problem. But again when inside `monorepos` `rollup` can't process through symlinks.
   

  The next option is dropping either `cjs` or `esm`.
  
  Dropping `cjs`
 - Dropping `cjs` is a safe bet at the moment, since `node14` can run modules natively. And browsers can run them and `bundlers` can use them. But one caveat, with `esm` is typescript. Typescript keeps complaining about import assertions with the package is not marked as `module` inside `package.json`. And with that and with the above mentioned bugs in which we can't bunch files. So, these two collide with each other.
 
 Dropping `esm`
- Riding along only with `cjs` actually solves a lot of things. But again now they effect browsers. Browsers can't run them `natively` but bundlers can use them, `node` can use them. So, again, we can't use it natively but you can still use via `bundlers`.

Using export maps
- Yes, using export maps, we can expose both `cjs` and `esm` in node understandable way. But what lacks, is we need to name the `esm` file with `.mjs` extension. If we are not planning to switch on the `type:module` flag. But, typescript can't do that, so need to reply on solving the import assertions too. That's brings to square one of prolbems with `esbuild` and `rollup` with monorepos.

So, all in all there a multiple problems in moving a monorepo structure along with typescript to full `esm` at the moment. The root cause of all the mishap is when we want to transpile to `es5` and push to `npm`. With `es6` we are quite good and esbuild with `tsc --noEmit` does the job for us.

JayaKrishnaNamburu:
https://github.com/webpack-contrib/terser-webpack-plugin/issues/410

PR #590, opened 2021-07-13
fix(vercel-publisher): Conditionally import crypto module depending on the environment

### LINKED ISSUE #562: Move from Vercel inlineFile.
Because inline deploy on Vercel is restricted to only 10MB we should rewrite the deploy functionality with uploading each file separately and the making a new deploy by providing the SHA1 of those files.

From Vercel docs:

> Uploading each file and then creating the deployment with files using SHA1 strings is especially useful when uploading many, or big, files

More info here: [https://vercel.com/docs/api#endpoints/deployments/upload-deployment-files](https://vercel.com/docs/api#endpoints/deployments/upload-deployment-files)

JayaKrishnaNamburu:
fixes #562


packages/teleport-publisher-vercel/src/crypto-node.js
```diff
@@ -0,0 +1,8 @@
+
+const { createHash } = require('crypto')
+
+export const encryptFileData = (fileData) => {
+    const hash = createHash('SHA1')
+    hash.update(fileData)
```
Utwo: I think this can be a one liner. 
`return createHash('SHA1').update(fileData).digest()`

packages/teleport-publisher-vercel/src/crypto.js
```diff
@@ -0,0 +1,4 @@
+export const encryptFileData = async (fileData) =>  {
+    const digest = await crypto.subtle.digest('SHA-1', fileData)
```
Utwo: We can do direct return here like:
`return crypto.subtle.digest('SHA-1', fileData)`

And if we target browsers, shouldn't we use `window.crypto.` ?

Utwo (changes requested):

PR #631, opened 2021-12-22
feat(github gateway): add github extra integration functionalities

ionutpasca:


packages/teleport-github-gateway/src/github-instance.ts
```diff
@@ -171,7 +219,7 @@ export default class GithubInstance {
         : await repo.createBlob(file.content)
 
     return {
-      sha: data.sha,
+      sha: file.status === 'deleted' ? null : data.sha,
```
JayaKrishnaNamburu, line 222: @ionutpasca we don't have `status` field in `file` object. You can either extend it here, if it's only needed in GitHub pacakage. Since, we don't use them in generators.

JayaKrishnaNamburu: Or you can check with `file.content` something ? 
```js
!file.content 
```
something of that sort if it works

ionutpasca: We actually have `status` JK. I added it. we cannot use your proposal, because the file content can be empty by desire. We need to have a specific implementation for when we want to actually remove a file, different than the one when we want to remove the content

JayaKrishnaNamburu: Ok, sounds good then. Can we update the types for GeneratedFolder and FeneratedFile to add optional status field in teleport-types. To pass this one

ionutpasca: My bad, didn't see that the build failed. Fixed now

PR #649, opened 2022-02-22
(default-func): Add void func as default handler for `propDefinitions` of type `func`

### LINKED ISSUE #638: Assign a default function instead of a string for PropTypes.Function in code-generation for react-components
```jsx
import React from 'react'

import PropTypes from 'prop-types'

import projectStyles from '../style.module.css'
import styles from './my-button.module.css'

const MyButton = (props) => {
  return (
    <div className={styles['container']}>
      <button onClick={props.doSomething} className={projectStyles['button']}>
        {props.MyButton}
      </button>
    </div>
  )
}

MyButton.defaultProps = {
  doSomething: '() => {}',
  MyButton: 'Button',
}

MyButton.propTypes = {
  doSomething: PropTypes.func,
  MyButton: PropTypes.string,
}

export default MyButton
```

This piece of code, definitely breaks the app. When props.doSomething is not passed and the `click` is simulated.

JayaKrishnaNamburu:
fixes #638

### DIFF
--- a/examples/uidl-samples/project.json
+++ b/examples/uidl-samples/project.json
@@ -992,6 +992,12 @@
     },
     "ContactForm": {
       "name": "ContactForm",
+      "propDefinitions": {
+        "isClicked": {
+          "type": "func",
+          "defaultValue": "() => {}"
+        }
+      },
       "node": {
         "type": "element",
         "content": {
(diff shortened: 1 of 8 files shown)


Utwo (approved):
:fire:

PR #659, opened 2022-03-08
Add remote options for fetch

TCern:
Adds possibility to add the 2nd parameter of the fetch in `getImageBufferFromRemoteUrl`

### DIFF
--- a/packages/teleport-publisher-vercel/src/browser.ts
+++ b/packages/teleport-publisher-vercel/src/browser.ts
@@ -12,8 +12,8 @@ export const getSHA = async (image: Uint8Array | ArrayBuffer) => {
 export const getImageBufferFromase64 = (content: string) =>
   Uint8Array.from(atob(content), (c) => c.charCodeAt(0))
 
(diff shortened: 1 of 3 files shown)


packages/teleport-publisher-vercel/src/browser.ts
```diff
@@ -12,8 +12,8 @@ export const getSHA = async (image: Uint8Array | ArrayBuffer) => {
 export const getImageBufferFromase64 = (content: string) =>
   Uint8Array.from(atob(content), (c) => c.charCodeAt(0))
 
-export const getImageBufferFromRemoteUrl = async (content: string) => {
-  const response = await fetch(content)
+export const getImageBufferFromRemoteUrl = async (content: string, options?: RequestInit) => {
```
JayaKrishnaNamburu, line 15: Hey @TCern you need to do the same for nodejs too. Because, This curretnyl effects only in `browser`. You need to make same changes inside `node.ts` file too

Utwo (approved):
:fire:

PR #660, opened 2022-03-08
refactor(github-gateway): use octokit to create file blobs before committing the files

ionutpasca:
…ommit


JayaKrishnaNamburu (approved):
LGTM

PR #661, opened 2022-03-09
feat(chore): testing out turborepo for decreasing build times

JayaKrishnaNamburu:


JayaKrishnaNamburu:
Makes a good fit, once the watcher is also added. Building all packages everytime for a change is still a little overkill. Everything is cached still 😓 

Screenshot 2022-03-17 at 8 59 48 PM

PR #665, opened 2022-03-15
Fix socket hang

Utwo:
Implement async-sema for limiting the number of concurrent requests. 
Remove browser support for vercel package because there is too much overwhelm in running both versions and at this time we don't have a proper use-case for browser usage.
Use node-fetch with keepAlive set to true, instead of cross-fetch package.


packages/teleport-publisher-vercel/package.json
```diff
@@ -8,14 +8,6 @@
   "main": "dist/cjs/index.js",
   "module": "dist/esm/index.js",
   "types": "dist/cjs/index.d.ts",
-  "imports": {
```
JayaKrishnaNamburu, line 11: Can we keep this one @Utwo, although we don't have a use-case to run in it browser right away. Would be a great advantage when anyone using the package want to do that

JayaKrishnaNamburu: But yeah, it's little complicated to implement two completly diff ehaviours. Since, in node env we are using semaphores and stuff. But maybe we can `expose` a single func from both the `env` files.

So, node version can be complicated with semaphores and retries. But `browser` can be simple without any reties and stuff like that ?

Not mandatory to have it now though

JayaKrishnaNamburu (approved):

PR #672, opened 2022-04-04
Adding support for comp-styles and comp-overrides

JayaKrishnaNamburu:
- Changing react to export css by default
- All flavours change the comp-scoped styles to dashCase(CSS, CSS-Modules, StyledJSX) and cameCase(StyledJSX, StyledComponents).
- Global style sheets of Plain CSS and StyledJSX don't change names, just simply appends whatever coming from UIDL (studio).
- Adding support for component-variants we have component-overrides before. This adds another feature
- Next with CSSModules fix for project template
- Lots of refactoring and  fixes along the way


Utwo (approved):
:fire:

PR #683, opened 2022-05-16
Use webcrypto with async functions instead of old crypto API

Utwo:
Use webcrypto with async functions instead of old crypto API


JayaKrishnaNamburu:
@Utwo CI seems to fail, can you check once

Utwo:
Should work now. Thanks!

JayaKrishnaNamburu (approved):
LGTM 👍

PR #696, opened 2022-07-20
(refactor):Improve the support for Image component from next

### LINKED ISSUE #686: Using `Image` from `next/Image` all the time for all images.
Right now, we are using `next/image` only if the user specifies `width` and `height` alone without any additional attributes. This is because, `next/Image` is not completely pure component at the moment. Because it is bringing some side effects tot he layout with it. Since we can't pass the `styles` that users set into the component. 

https://github.com/teleporthq/teleport-code-generators/blob/development/packages/teleport-plugin-jsx-next-image/src/index.ts#L11

Here is the issue on the `NextJS` GitHub on the same topic
https://github.com/vercel/next.js/discussions/18312

Case 1
It is not working out of the box all the time, since we allow users to free-style on the properties they use for the images. But the `next/image` need to be used with some constraints
Usual - https://codesandbox.io/s/pf9mt4
next/Image - https://codesandbox.io/s/travel-agency-forked-h28f9p

JayaKrishnaNamburu:
fixes #686 

But, currently the new `Image` component doesn't work as expected with `styled-jsx`. Here is the issue https://github.com/vercel/next.js/issues/38822 Let's wait for some info on that


JayaKrishnaNamburu:
I came back looking at this, @mihaisampaleanu here is where we left the discussion to handle this. Currently `next/image` don't support `className` prop and apply them on the image. Instead they ask to adopt `css-modules`. So, we should decide if we move ahead and make `Next - CSS Modules` as default for the nextjs projects.

The only down side is, any legacy `custom-scripts` that are following `classnames` might break. Since `css-modules` hash the class-names for deployments. But, at the end this switch will be a huge perf gain. So, should we switch to it and give warning to old scripts users ?

PR #716, opened 2022-10-21
(feat): Adding support for raw attribute type

elizamaria:
Supports sending an attribute of type `raw` when we want to skip encoding an attribute. 

Handles attributes for html component as well -> wanting to send a `srcdoc` attribute for an iframe => avoid encoding


packages/teleport-plugin-common/src/utils/ast-utils.ts
```diff
@@ -174,6 +174,19 @@ export const addAttributeToJSXTag = (
...
+export const addRawAttributeToJSXTag = (
+  jsxNode: types.JSXElement,
+  attrName: string,
+  attrValue: UIDLRawValue,
+  t = types
+) => {
+  const attributeDefinition = t.jsxAttribute(
+    t.jsxIdentifier(attrName),
+    t.jsxExpressionContainer(convertValueToLiteral(attrValue.content))
+  )
```
JayaKrishnaNamburu, line 188: Here instead of `convertValueToLiteral` you can do something like this
```js
types.templateLiteral([types.templateElement({ raw: attrValue.content })], [])
```

So, this will generate string template literals. Which helps n better readability with lengthy strings

JayaKrishnaNamburu (changes requested):
All looks good, just added tests and i think we are good to merge 🚀

JayaKrishnaNamburu:
@all-contributors please add @elizamaria for code

PR #720, opened 2022-10-28
Feat/html node: Handle code embed node in generators

elizamaria:
`html-node` will be rendered using the dangerous-html package for all frameworks except Next.js where dangerous-html-content will be used
- new type of attribute is supported -> `raw` which escapes encoding
- update schemas in Angular to support custom elements


JayaKrishnaNamburu:
- [x] Stencil
- [x] Html
- [x] Gridsome
- [x] Gatsby
- [x] React
- [x] Preact
- [x] Angular
- [x] Nuxt
- [x] Vue
- [x] Next

Seems stencil generators are not being added properly. Needs to check the cause. If we fix, it works with all frameworks
[Edit]: Stencil is also fixed

packages/teleport-project-generator-vue/__tests__/end2end/index.ts
```diff
@@ -63,6 +64,7 @@ describe('Vue Project Generator', () => {
     expect(modalComponent.content).toContain(`import { Button } from 'antd'`)
     expect(viewsFolder.files[0].content).toContain(`<app-modal></app-modal>`)
     expect(viewsFolder.files[0].content).toContain(`import AppModal from '../components/modal'`)
+    expect(viewsFolder.files[0].content).toContain(`import 'dangerous-html'`)
```
JayaKrishnaNamburu: For vue, the import should be
```js
import DangerousHTML from 'dangerous-html/vue'
```

From the mapping from here https://github.com/teleporthq/teleport-code-generators/pull/720/files#diff-dd49a3c1b9b0a9d400aea689415c3da9508046bfd40fb870cffaf03dedb26ff1R12

JayaKrishnaNamburu (approved):
💯 🚀

PR #722, opened 2022-11-03
Fix build when project name contains paranthesis

elizamaria:
Issue: 
 1. save on local machine a vue project which has a parenthesis in the project name.
 2. try building the project
 
Actual: build fails

Related issue ->  https://github.com/vuejs/vue-cli/issues/7043

### DIFF
--- a/packages/teleport-project-generator-vue/src/project-template.ts
+++ b/packages/teleport-project-generator-vue/src/project-template.ts
@@ -39,7 +39,9 @@ export default {
     {
       name: 'vue.config',
       fileType: 'js',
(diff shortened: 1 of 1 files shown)


JayaKrishnaNamburu:
Is this only a `vue` framework related projects @elizamaria. Or else, it's better to fix in the project name generation itself

elizamaria:
> Is this only a `vue` framework related projects @elizamaria. Or else, it's better to fix in the project name generation itself

Yes yes

JayaKrishnaNamburu (approved):
💯

PR #723, opened 2022-11-04
(refactor): Wrap `html-node` with a div with styles lifted to the paret

elizamaria:
Currently the html-node doesn't consider any style we're sending. In order to do that, we wrap it into a div element that will have the node content (styles, events, attributes) it receives on the html-node and it will render it as a child node.


elizamaria:
Should also update the dangerous-html package after it supports display contents

JayaKrishnaNamburu (dismissed):
All good, just update the package version number to `0.1.11`

JayaKrishnaNamburu (approved):
💯

PR #725, opened 2022-11-29
feature/nested-url-support

vylly:
WIP: add support for nested url (folder/mypage) => this includes css files path management / component import management / folder generation

### DIFF
--- a/packages/teleport-plugin-css/src/index.ts
+++ b/packages/teleport-plugin-css/src/index.ts
@@ -321,7 +321,8 @@ const createCSSPlugin: ComponentPluginFactory<CSSPluginConfig> = (config) => {
        * The name of the file is either in the meta of the component generator
        * or we fallback to the name of the component
        */
-      const cssFileName = UIDLUtils.getStyleFileName(uidl)
+      let cssFileName = UIDLUtils.getStyleFileName(uidl)
+      cssFileName = cssFileName.split('/').pop()
 
       if (declareDependency === 'decorator' && componentDecoratorChunk) {
         const decoratorAST = componentDecoratorChunk.content
--- a/packages/teleport-project-generator-html/src/index.ts
+++ b/packages/teleport-project-generator-html/src/index.ts
@@ -20,6 +20,9 @@ const createHTMLProjectGenerator = (config?: { individualEntyFile: boolean }) =>
     pages: {
       generator: createHTMLComponentGenerator,
       path: [''],
+      options: {
+        useFileNameForNavigation: true,
+      },
     },
     static: {
       prefix: 'public',
--- a/packages/teleport-project-generator-html/src/index.ts
+++ b/packages/teleport-project-generator-html/src/index.ts
@@ -25,7 +25,7 @@ const createHTMLProjectGenerator = (config?: { individualEntyFile: boolean }) =>
       },
     },
     static: {
-      prefix: 'public',
+      prefix: '/public',
       path: ['public'],
     },
     projectStyleSheet: {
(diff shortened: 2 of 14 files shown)


JayaKrishnaNamburu (changes requested):
This is a little bit of bigger refactor than it looks. You need to make the changes here
https://github.com/teleporthq/teleport-code-generators/blob/development/packages/teleport-project-generator/src/utils.ts#L200

For now, we are actually not relying on `routerLink` to calculate the folder path. We depend on `path` inside `ProjectStrategy`. Since the requirement is depend on `routerLink` from now on. We need to take that into consideration while calculating the output options here.
https://github.com/teleporthq/teleport-code-generators/blob/development/packages/teleport-project-generator/src/utils.ts#L182

So, that path will be used back here automatically and written to the output
https://github.com/teleporthq/teleport-code-generators/blob/development/packages/teleport-project-generator/src/index.ts#L230

So, the path checks happens only once and be synced across all plugins and generators.

JayaKrishnaNamburu:
The main reason that we need to take `routeLink` into consideration now. Because, many frameworks needs the pages in that respective paths for the routes to work.

packages/teleport-plugin-css/src/index.ts
```diff
@@ -321,7 +321,8 @@ const createCSSPlugin: ComponentPluginFactory<CSSPluginConfig> = (config) => {
        * The name of the file is either in the meta of the component generator
        * or we fallback to the name of the component
        */
-      const cssFileName = UIDLUtils.getStyleFileName(uidl)
+      let cssFileName = UIDLUtils.getStyleFileName(uidl)
+      cssFileName = cssFileName.split('/').pop()
```
JayaKrishnaNamburu, line 325: Please move this check inside the `getStyleFileName` so we don't need to make this check every time we try to get the file name

packages/teleport-project-generator-html/src/index.ts
```diff
@@ -20,9 +20,12 @@ const createHTMLProjectGenerator = (config?: { individualEntyFile: boolean }) =>
     pages: {
       generator: createHTMLComponentGenerator,
       path: [''],
+      options: {
+        useFileNameForNavigation: true,
+      },
     },
     static: {
-      prefix: 'public',
+      prefix: '/public',
```
JayaKrishnaNamburu, line 28: Can you check this, we don't need to add `/` the server takes care of it i think

JayaKrishnaNamburu (changes requested):

packages/teleport-project-generator-html/src/index.ts
vylly: I checked, for html specifically I needed to add this, if not the pages in subfolders will have relative asset paths to the public folder and it will not be valid
see: https://codesandbox.io/s/rctgk4 (with prefix: 'public')
and this: https://codesandbox.io/s/0rk9bp (with prefix: '/public')

I couldn't find another workaround so far

packages/teleport-plugin-css/src/index.ts
vylly: If you are talking about this `cssFileName = cssFileName.split('/').pop()` I cannot move it to the `getStyleFileName` as I only need his pop to happen when getting the filname to use it as a styleurl
If I insert this pop in the `getStyleFileName` function it breaks the css files generation as it uses this function as well but needs to be aware of the "/" contained in the filname in this case

JayaKrishnaNamburu (approved):
Cool, nice work @vylly

PR #730, opened 2022-12-24
feat: Upgrade deps and change scripts to use turbo

JayaKrishnaNamburu:
@elizamaria i just reviewed one of the old PR using turbo. You can try and merge if it seems to work for you 😄. This helps seep up only the local development and not the GitHub CI's. Unless we connect the `turbo` with vercel CI for cache. But nevertheless it will speed up local builds atleast.


packages/teleport-plugin-common/src/builders/ast-builders.ts
```diff
@@ -182,6 +182,20 @@ export const createStateChangeStatement = (statement: UIDLStateModifierEvent, t
...
+  const result = parse('<style>{`' + str.trim() + '`}</style>', {
+    sourceType: 'module',
+    plugins: ['jsx'],
+  })
+  const expressionConntainer = (
+    (result.program.body[0] as types.ExpressionStatement).expression as types.JSXElement
+  ).children[0] as types.JSXExpressionContainer
+  return expressionConntainer
+}
+
```
JayaKrishnaNamburu, line 198: The upgrade in typescript has really found some deep issues that we didn't notice before. One of it is the following. We used to build strings like this
```js
.component-with-dyamic {
  width: ${props.width}
}
```
As a string and then pass the string to `generateStyledJSXTag` ad convert it to template literal. Building template literals with dynamic values inside is very tricky. So instead of re-working the whole generation process. 

I used `@babel/parser` and then used the parser to parse this string, and take the ast ad inject it back to the code generators AST. It's more like string -> AST -> string conversion.

Not the best, but creating direct AST's for template literals with strings is very very tricky. So, this is the easiest one

packages/teleport-publisher-vercel/package.json
```diff
@@ -30,7 +30,7 @@
     "node-fetch": "^2.6.1"
   },
   "devDependencies": {
-    "@types/async-retry": "1.4.3",
+    "@types/async-retry": "^1.3.3",
```
JayaKrishnaNamburu, line 33: We are using `async-retry@1.3.3` but loading types of `1.4.3`. So basically types are not matching and another typescript error catch

PR #739, opened 2023-02-09
Small fix on the lottie pkg name for html project

elizamaria:
npm package name is wrong, it should be @lottiefiles/lottie-player

### DIFF
--- a/packages/teleport-component-generator-html/src/plain-html-mapping.ts
+++ b/packages/teleport-component-generator-html/src/plain-html-mapping.ts
@@ -27,7 +27,7 @@ export const PlainHTMLMapping: Mapping = {
       elementType: 'lottie-player',
       dependency: {
         type: 'package',
-        path: 'lottie-player',
+        path: '@lottiefiles/lottie-player',
         version: '1.5.7',
         meta: {
           importJustPath: true,


packages/teleport-component-generator-html/src/plain-html-mapping.ts
```diff
@@ -27,7 +27,7 @@ export const PlainHTMLMapping: Mapping = {
       elementType: 'lottie-player',
       dependency: {
         type: 'package',
-        path: 'lottie-player',
+        path: '@lottiefiles/lottie-player',
```
JayaKrishnaNamburu, line 30: But @elizamaria this shouldn't effect anything right 🤔. Because `html-generators` basically takes `importAlias` and creates `script` tags ?? Was it creating any issues ?

elizamaria: Yes, it was breaking when exporting the project in codesandbox.

elizamaria: Thanks for handling the fix, JK

JayaKrishnaNamburu: Ahh, codesandbox tries to resolve it. I think the best way here is. But, fun fact it. We don't use those any ways. Since there is no build step involved.

JayaKrishnaNamburu: Can you check this one too @elizamaria  -> https://github.com/teleporthq/teleport-code-generators/pull/735

PR #740, opened 2023-02-14
Fix/kill unmaintained generators

andreiTnu:
Removed
* gridsome 
* gatsby
* preact 
* stencil


JayaKrishnaNamburu:
I would still retain `preact` 😅. It's on par with `react` and we don't do any additional changes for it 😁

PR #748, opened 2023-03-22
Add support for fetch requests and Nextjs static paths and props

ionutpasca:


ionutpasca:
**Make sure to make this one before** -> https://github.com/teleporthq/teleport-code-generators/pull/749

elizamaria:
Looks like it's good to go. Should check with @vylly the style/assets support for the nested pages.

packages/teleport-project-plugin-contexts/src/next.ts
```diff
@@ -0,0 +1,51 @@
...
+    const contextContent = prettierJS({
+      [FileType.JS]: `
+import { createContext, useContext } from 'react'
+
+export const ${contextName} = createContext(null)
+
+export const use${contextName} = () => {
+  const context = useContext(${contextName})
+  if (!context) {
+    throw new Error('use${contextName} must be used within a ${contextName}')
```
vladnicula, line 27:
```suggestion
    throw new Error('use${contextName} must be used within a ${contextName}Provider')
```

packages/teleport-plugin-next-context/src/index.ts
```diff
@@ -0,0 +1,95 @@
+import {
```
vladnicula, line 1: this file is for context providers, not context declaration or read. I was expecting a context read/access not provision in this file. Maybe naming could be better.

examples/uidl-samples/cms-project.json
```diff
@@ -0,0 +1,3684 @@
...
+                  "name": "BookEntity",
+                  "valuePath": [],
+                  "itemValuePath": [
+                    "fields"
+                  ]
+                },
+                "resourceMappers": [
+                  {
+                    "name": "resolveContentfulResponse",
+                    "resource": {
```
JayaKrishnaNamburu, line 551: Can this be changed to `dependency` to keep it consistent with UIDLElemetNode. And second if we can use `importDefinitions` and just make it a reference here. Would also be a option
https://github.com/teleporthq/teleport-code-generators/blob/53c7b8a9ffbe8e18341da464580e08f5292c3f00/examples/uidl-samples/react-project.json#LL181C1-L181C1

examples/uidl-samples/cms-project.json
```diff
@@ -0,0 +1,3684 @@
...
+                      "meta": {
+                        "originalName": "resolveContentfulResponse"
+                      }
+                    }
+                  },
+                  {
+                    "name": "normalizeContent",
+                    "resource": {
+                      "type": "package",
+                      "version": "git+https://github.com/ionutpasca/contentful-normaliser.git",
```
JayaKrishnaNamburu, line 771: This can be moved into `teleporthq` organisation. And we can release it. And make the version `latest`. So, the deployments always get the latest normaliser

examples/uidl-samples/cms-project.json
```diff
@@ -0,0 +1,3684 @@
...
+            }
+          ]
+        }
+      },
+      "name": "BlogPostDetailsContent",
+      "styleSetDefinitions": {}
+    }
+  },
+  "resources": {
+    "rootFolder": "pages/api",
```
JayaKrishnaNamburu, line 3621: We can make this framework's strategy dependent instead of `pages/api`.

examples/uidl-samples/cms-project.json
```diff
@@ -0,0 +1,3684 @@
...
+    "items": {
+      "7816efdd-2f6b-46f5-b5c6-a4336510c92f": {
+        "id": "7816efdd-2f6b-46f5-b5c6-a4336510c92f",
+        "name": "Home",
+        "baseUrl": {
+          "type": "env",
+          "value": "CMS_URL"
+        },
+        "authToken": {
```
JayaKrishnaNamburu, line 3630: We can split this into `headers` and `params`. And each param can have a param type. Like `url_param` and `obj` for post data. Something like that. We can switch the `{ method: 'POST' }` if the params preset as `obj` something like that.

examples/uidl-samples/cms-project.json
```diff
@@ -0,0 +1,3684 @@
...
+      "bea417dd-a288-41cc-9c75-df5fe292ba2a": {
+        "name": "Book",
+        "fileName": "Book"
+      },
+      "536f3717-684e-4d58-816f-29975cc9bc93": {
+        "name": "BlogPost",
+        "fileName": "BlogPost"
+      }
+    }
+  }
```
JayaKrishnaNamburu, line 3683: Should we abstract out the whole `params` and `url_params`. We can write a `lib` that takes, `path`, `params` andmake the fetch calls. Instead of generating every time. Something like
``
import fetcher from '@teleporthq/fetcher'

const res = fetcher(path, params)
``

So, just params ad have param type, `token`, `post`, `url_param` something like that

JayaKrishnaNamburu: Just like `axios` but with opinionated for our requirements ?

packages/teleport-plugin-common/src/builders/style-builders.ts
```diff
@@ -146,11 +157,19 @@ export const generateStylesFromStyleSetDefinitions = (
       if (styleRef.type === 'element-state') {
         if (type === 'reusable-component-style-map') {
           cssMap.unshift(
-            createCSSClassWithSelector(name, `&:${styleRef.meta.state}`, collecedMediaStyles)
+            createCSSClassWithSelector(
+              name,
+              `&${subselectors || ''}:${styleRef.meta.state}`,
```
JayaKrishnaNamburu, line 162: Can we try using this approach for selectors in react-jss
https://cssinjs.org/jss-plugin-nested?v=v10.6.0

JayaKrishnaNamburu (commented):
Some improvements we can do, we can extract out something like `fetcher` and some framework specific things. We can extend the project strategy with `contexts` and `resources`. So, they will be handled right with the project generators. Without adding any plugins. Or maybe add those plugins by default to the project-generators. So, by default they will start working.

JayaKrishnaNamburu:
Something like this

```js
const generator = createProjectGenerator({
    id: 'teleport-project-next',
    style: ReactStyleVariation.StyledJSX,
    components: {
      ....
    },
    pages: {
      .....
    },
    projectStyleSheet: {
      .....
    },
    entry: {
     .....
    },
    framework: {
      config: {
        ....
      },
      externalStyles: {
        ....
      },
    },
    static: {
     ...
    },
   cms: {
    ....
   }
  })
```

Something like this, so each framework can use it's own plugins to read the paths and things like that.

examples/uidl-samples/cms-project.json
ionutpasca: Yes, we can do that

examples/uidl-samples/cms-project.json
ionutpasca: Yes...this was only to ship this functionality faster. We can use any package from anywhere :D

examples/uidl-samples/cms-project.json
ionutpasca: Yes...this is a good idea

examples/uidl-samples/cms-project.json
ionutpasca: Yes... we had a discussion about this. We could have a `headers` property inside this object and `authToken` would be inside of it. It remained like this for a faster shipment :)

examples/uidl-samples/cms-project.json
ionutpasca: This is an option as well, we only had an internal discussion about it. We could do it, but it would mean that the implementation is hidden from the code. We need to decide if this is what we want to do or not. As a developer, I would like to see what and how the fetch request is done, not to be hidden from me; this is a subjective opinion

ionutpasca:
>

ionutpasca:
Ideally, we should not have any CMS related stuff inside the UIDL. The fact that we added a CMSListItem and a CMSItem node inside the the UIDL was only to speed up the development. We added the `resource` entity to be agnostic from any external dependency that one might use inside the project. we basically needed a `fetch` integration, that is currently used with the CMS, but it could be used in the same time with any external API.

I'm not really sure what you would add inside the `cmd` property that you mentioned above:)

JayaKrishnaNamburu:
Hey, i got it. Just bouncing some ideas, if we want the CMS too be inside the UIDL. But, `cms` inside strategy will be allow the pre-defined setups that are needed. Like deps, fetchers, api locations etc. Anything that the framework needs to fetch the data.

PR #749, opened 2023-03-23
Add support for pagination on static paths/props

ionutpasca:


examples/uidl-samples/cms-project.json
```diff
@@ -0,0 +1,1324 @@
...
+                },
+                "route": {
+                  "type": "static",
+                  "value": "api/books"
+                }
+              }
+            }
+          },
+          {
+            "value": "books/Books1",
```
elizamaria: This should be books/Books

examples/uidl-samples/cms-project.json
```diff
@@ -0,0 +1,1324 @@
...
+                {
+                  "property": "og:title",
+                  "content": "Soggy Insubstantial Octopus"
+                }
+              ]
+            },
+            "pageOptions": {}
+          },
+          {
+            "value": "/books/Books",
```
elizamaria: This should be books/Book1

examples/uidl-samples/cms-project.json
```diff
@@ -0,0 +1,1324 @@
...
+                "authToken": {
+                  "type": "env",
+                  "value": "CMS_ACCESS_TOKEN"
+                },
+                "urlParams": {
+                  "populate": {
+                    "type": "static",
+                    "content": "*"
+                  },
+                  "pagination": {
```
elizamaria, line 683: Can't we skip sending this? We already send all pagination related info in the "pagination" object, and we know the pageSize

elizamaria (approved):
Good go to

examples/uidl-samples/cms-project.json
ionutpasca: Make a change

examples/uidl-samples/cms-project.json
ionutpasca: Made a change

examples/uidl-samples/cms-project.json
ionutpasca: We know the pagination stuff, but this is a more general implementation. This is inside the `initialPropsResource - urlParams`, that can be anything...  Whenever we are building the initial props, we don't really want to use information from outside of this scope. Does it make sense?

PR #751, opened 2023-03-23
(feat): Support for transforming code-embed into external files or to jsx and html respecitvely

JayaKrishnaNamburu:
We have two new plugins in this
- `teleport-project-plugin-external-embed` this will externalise the code into a separate files. So, the code looks clean and the embed is loaded using a fetch call dynamically. One huge advantage is, embeds are written by users. So if there is any syntax errors, the whole code-generation pipeline breaks. Which happens in complex templates already. Take this as example https://play.teleporthq.io/projects/togthr-template-m453i9. We should make this a default behaviour for `vue` and `angular`
- Next we have `teleport-project-plugin-parse-embed`, this is the main one. This parses the code written by users and converts it into `jsx` or `html` w.r.t to the end target. So, basically the embeds that users are writing goes as a proper code generated. Unless if it is `script` tag for `jsx`. Because `script` is not supported in jsx by default.

One major improvement is, using the `teleport-project-plugin-parse-embed` makes the script embeds to look like `NextJS`. Example
In next it looks like `import Script from 'next/script` and in react it looks like `import Script from 'dangerous-html`. Just the change in imports, the whole code looks exactly the same.


JayaKrishnaNamburu:
Hey @elizamaria this is fixed now
https://github.com/JayaKrishnaNamburu/hast-util-to-jsx-fork/commit/5bea3d23bf3ab942355b1ca9d7e5b6a9eb5d6a46#diff-e727e4bdf3657fd1d798edcd6b099d6e092f8573cba266154583a746bba0f346R45

Turns out, the script tag without content is not handled before. It should be good now 👍

PR #753, opened 2023-03-28
fix/default-props-containing-assets-fix

vylly:
fixing default props path if default props is an asset


JayaKrishnaNamburu (changes requested):
Since you started working on this, @vylly can you add the same for `stateDefinitions` too ?

JayaKrishnaNamburu (approved):

JayaKrishnaNamburu:
@all-contributors add @vylly for code

PR #757, opened 2023-03-30
fix/link-path-using-prefix-assets-path

vylly:
Local links are using the prefixAssetsPath function and they start with a '/' they be considered as asset and then prefixed with public folder and / or by asset identifier if there is any
An idea would be, in the case the node being resolved is a link to send an empty object instead of the assets options to not append anything before the local links

### DIFF
--- a/packages/teleport-uidl-resolver/src/utils.ts
+++ b/packages/teleport-uidl-resolver/src/utils.ts
@@ -138,7 +138,9 @@ export const resolveElement = (element: UIDLElement, options: GeneratorOptions) 
       if (attrValue.type === 'static' && typeof attrValue.content === 'string') {
         originalElement.attrs[attrKey].content = UIDLUtils.prefixAssetsPath(
           attrValue.content,
-          options.assets
+          originalElement.elementType === 'Link' || originalElement.elementType === 'a'
+            ? {}
+            : options.assets
         )
       }
     })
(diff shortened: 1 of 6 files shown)


packages/teleport-uidl-resolver/src/utils.ts
```diff
@@ -138,7 +138,9 @@ export const resolveElement = (element: UIDLElement, options: GeneratorOptions)
       if (attrValue.type === 'static' && typeof attrValue.content === 'string') {
         originalElement.attrs[attrKey].content = UIDLUtils.prefixAssetsPath(
           attrValue.content,
-          options.assets
+          originalElement.elementType === 'Link' || originalElement.elementType === 'a'
```
JayaKrishnaNamburu: This acts kinda of hack, let's come up with some solution. But ideally a link should me missing from assets mapping. So, we should never run into this issue 🤔

JayaKrishnaNamburu (changes requested):

packages/teleport-uidl-resolver/src/utils.ts
vylly: The problem is not about the assets mapping (the link is not in the asset mapping), the issue is either in the fact that local links are accessing this prefixing function or in the prefixing function itself 
`if (!mappings[assetName]) {
    if (!identifier) {
      return [prefix, assetName].join('/')
    }
    return [prefix, identifier, assetName].join('/')
  }`
  
  The problem is that if we have an identifier (or a prefix or both) defined in the options.assets (which is necessary for assets) it will be appended to a local link even though it doesn't make sense to use assets related properties on it

JayaKrishnaNamburu (approved):

PR #760, opened 2023-04-05
Add CMS list+item nodes support

ionutpasca:


examples/uidl-samples/cms-project.json
```diff
@@ -990,11 +1015,132 @@
...
+                            "referenceType": "prop",
+                            "id": "BooksEntity.attributes"
+                          }
+                        }
+                      },
+                      "children": [
+                        {
+                          "type": "element",
+                          "content": {
+                            "elementType": "fragment",
```
elizamaria, line 1039: Because of this I think the exported code looks like this:
` <BooksContext.Provider value={props.BooksEntity.attributes}>
        <fragment>
          <BooksDetailsContent></BooksDetailsContent> 
        </fragment>
      </BooksContext.Provider>`

Maybe not send anything or just replace with <> or React.Fragment

elizamaria: Discussed: will be handled in another PR

PR #765, opened 2023-04-19
Feat/dynamic links support for cms details pages

vladnicula:
use `Array.from($$('.check-annotation')).forEach((item) => item.style.display = 'none')` to hide the annoying code coverage messages.

see https://linear.app/teleporthq/issue/THQ-1963/add-a-resolution-of-the-new-cms-details-links-when-mapping-domain-data


package.json
```diff
@@ -32,7 +32,7 @@
     }
   },
   "engines": {
-    "node": "16.x"
```
vladnicula, line 35: I have node 18 and cannot run the generators. What gives?

packages/teleport-plugin-common/src/node-handlers/node-to-jsx/utils.ts
```diff
@@ -147,20 +147,33 @@ export const createDynamicValueExpression = (
   params?: JSXGenerationParams
 ) => {
   const { projectContexts = {} } = params || {}
-  const refType = identifier.content.referenceType as 'prop' | 'state' | 'local' | 'ctx'
```
vladnicula, line 150: this type cast does not guarantee we don't call this function without the other types of dynamic values. Better throw errors instead of assuming we know how to use the function.

PR #769, opened 2023-04-20
Classes with meta tag for subselector support

TudorCe:
Added `meta` tag to reusable project style sheets. This approach can also help get rid of the state condition.

Alternatively, if this approach seems bad or has issues, the same approach with the state conditions can be used.


JayaKrishnaNamburu:
Hey @TudorCe just checking around. What's the "state condition" issue ? are you looking for not to change the class-names or something ?

TudorCe:
Hey @JayaKrishnaNamburu , there are no issues with the state condition. This change aims to support classes with subselectors (so, stuff like `.classname div > p`) in some way. I've added these changes to how classes are generated as one way of doing it, though the same approach as used for states can be used here as well.

I'm really not sure which approach would be better, both have +s and -s :D let me know if you have any concerns

JayaKrishnaNamburu:
We have something like `conditions`. So, i think that would be the proper way to handle it. Let's see

Issue #766, opened 2023-04-19
`lottieplayer` is being added to all next projects by default


JayaKrishnaNamburu:
We should refactor the the additation of `lottieplayer` to all next projects. We are currently adding it by default and importing it. Which just increases the bundle size. We should add for only the projects that are using lottie files. At the same time, check for all other frameworks too.

PR #777, opened 2023-04-22
(fix): Add lottie player only when used in next projects

JayaKrishnaNamburu:
This fixes build size for all NextJS projects. I will cross check for all other frameworks too. `lottie` player is little heavy. Right now, it's being added into all projects. 

Screenshot 2023-04-22 at 20 30 12
Screenshot 2023-04-22 at 20 30 46

### DIFF
--- a/examples/uidl-samples/project.json
+++ b/examples/uidl-samples/project.json
@@ -819,6 +819,26 @@
                                                     {
                                                         "type": "element",
                                                         "content": {
+                                                            "elementType": "lottie-node",
+                                                            "style": {
+                                                                "width": "200px",
+                                                                "height": "200px"
+                                                            },
+                                                            "attrs": {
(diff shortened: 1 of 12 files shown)


JayaKrishnaNamburu:
The main reason was, `lottie-player` is something imported and registered in DOM. So, the bundlers can't perform efficient tree-shaking and so it's being added to bundle all the time. Fixes #766

JayaKrishnaNamburu:
@elizamaria any specific issues that we are not using `@lottiefiles/react-lottie-player` with NextJS. Because currently what i am seeing is, the styles are also not propagating to the web-component. So, need to fix the styling with `lottie-node` in NextJS as we are using web-components.

Anyways, either way we need to fix it. Since, the `styled-jsx` styles are not propagating even if we use with `react` version of it.

PR #785, opened 2023-05-03
Feat/markdown richtext node

elizamaria:
Handles markdown-node element type
Renders markdown content as HTML content using the external package


packages/teleport-component-generator-react/src/react-mapping.ts
```diff
@@ -21,6 +21,14 @@ export const ReactMapping: Mapping = {
         version: '2.2.1',
       },
     },
+    'markdown-node': {
+      elementType: 'Markdown',
```
JayaKrishnaNamburu, line 25:
```js
elementType: 'markdown'
```

Please 😀

JayaKrishnaNamburu: Should we build something like `dangerous-html` for this use-case too ?

JayaKrishnaNamburu:
I think we can improve this a lot, instead of rendering markdown always to jsx at runtimes. We can use this
https://github.com/remarkjs/remark-rehype#remark-rehype
https://unifiedjs.com/explore/package/mdast-util-from-markdown/#frommarkdownvalue-encoding-options

And convert the markdown to `jsx` and `html` respectively. Users can write markdown, but there is no runtime on markdown to jsx. Because the way these values are changes only on the `studio` and nothing post build so this is most promising 😁 

Here is a example https://github.com/teleporthq/teleport-code-generators/pull/751/files#diff-d0233a558de7f5795a6759f3a72fb3b11f20aedf5fa077f54c32b3f34989a0aaR38

packages/teleport-component-generator-react/src/react-mapping.ts
elizamaria: This is how the element will look in jsx code, why with small letter? No need to build something new, this one works well.

JayaKrishnaNamburu: Usually we go by, `elementType` with small case and then `semanticType` with the actual import or component name.

Yes, this works out of the box with `react` i am referring to all other frameworks and we can just do the. `mdx-to-jsx` at code-generation time. Without any runtime libraries 👍

elizamaria:
> I think we can improve this a lot, instead of rendering markdown always to jsx at runtimes. We can use this https://github.com/remarkjs/remark-rehype#remark-rehype https://unifiedjs.com/explore/package/mdast-util-from-markdown/#frommarkdownvalue-encoding-options
> 
> And convert the markdown to `jsx` and `html` respectively. Users can write markdown, but there is no runtime on markdown to jsx. Because the way these values are changes only on the `studio` and nothing post build so this is most promising 😁
> 
> Here is a example https://github.com/teleporthq/teleport-code-generators/pull/751/files#diff-d0233a558de7f5795a6759f3a72fb3b11f20aedf5fa077f54c32b3f34989a0aaR38

Content will change after build, because this content can be updated from the CMS at any moment and the changes should be visible.

JayaKrishnaNamburu:
ah, ok it's linked with CMS. Got it 👍

packages/teleport-component-generator-react/src/react-mapping.ts
elizamaria: for now we're focusing on publishing with next.js the things related to cms, if we will intend to support other frameworks we will indeed need to find another way. I will try with semnaticType

elizamaria: JK I see we're using elementType for html-node, lottie-node as well, with uppercase, so this seems to be the common behavior. Also, if I'm having elementType together with semnaticType, elementType will be used in the generated jsx. I suggest leaving it as it is for now, what do you think?

JayaKrishnaNamburu: yeah sure sure, no problem 👍

PR #787, opened 2023-05-04
(refactor): Refactor better fetch branch and UIDL

JayaKrishnaNamburu:


packages/teleport-types/src/uidl.ts
```diff
@@ -26,14 +26,25 @@ export interface ContextsUIDL {
...
+export interface UIDLResourceItem {
   name: string
+  headers?: Record<string, UIDLStaticValue | UIDLENVValue>
+  path: {
+    baseUrl: UIDLStaticValue | UIDLENVValue
+    route: UIDLStaticValue
+  }
+  params?: Record<string, UIDLStaticValue>
+  method?: 'GET' | 'POST'
+  body?: Record<string, UIDLStaticValue>
```
JayaKrishnaNamburu, line 52: @ionutpasca @vladnicula let's discuss more on this, we can generate resource files separately.  And map those resource fetchers from `initialProps` and `getStaticProps` and re-use them. Works as a base for generating fetchers in all other frameworks too

JayaKrishnaNamburu:
Every resource end-point get's it's own module. And the pages import them and then used directly used inside the `getStaticProps`  function.


```js
import entries from '../resources/entries'

import React, { useEffect } from 'react'
import Head from 'next/head'

import DangerousHTML from 'dangerous-html/react'
import resolveContentfulResponse from 'contentful-resolve-response'
import { normalize } from '@teleporthq/teleport-cms-mappers/contentful'
......
.....
....


export async function getStaticProps(context) {
  const mappedData = normalize(resolveContentfulResponse(response))
  return {
    props: {
      BookEntity: mappedData,
    },
    revalidate: 1,
  }
}
```

So, we can use the resource module by importing framework specific way.

JayaKrishnaNamburu:
I even have this strange idea, that maybe we should move the normalisers into `resources`. So, when the resource is requested. It can always be normalized without importing in every component. What say ??

Because they are resource oriented instead of the component props side. So every resource can have a new attr like `mappers` instead of adding them on `initialProps`

```js
export default async function () {
  const data = await fetch(
    `${process.env.CMS_URL}/entries?${new URLSearchParams({
      content_type: 'author',
      limit: '10',
    })}`,
    {
      method: 'GET',
      headers: {
        'Content-Type': 'application/json',
        Authorization: `Bearer ${process.env.CMS_ACCESS_TOKEN}`,
      },
    }
  )
  const response = await data.json()
  return normalize(contentfulNormalize(response))
}

```

JayaKrishnaNamburu:
Here is the new change
```js
import resolveContentfulResponse from 'contentful-resolve-response'
import { normalize } from '@teleporthq/teleport-cms-mappers/contentful'

export default async function () {
  const data = await fetch(
    `${process.env.CMS_URL}/entries?${new URLSearchParams({
      content_type: 'author',
      limit: '10',
    })}`,
    {
      method: 'GET',
      headers: {
        'Content-Type': 'application/json',
        Authorization: `Bearer ${process.env.CMS_ACCESS_TOKEN}`,
      },
    }
  )
  const response = await data.json()
  return normalize(resolveContentfulResponse(response))
}

```


```json
"6b5f14bd-1b8f-4b6a-b592-9f2e4e56a475": {
        "id": "6b5f14bd-1b8f-4b6a-b592-9f2e4e56a475",
        "name": "Entries",
        "mappers": {
          "resolveContentfulResponse": {
            "type": "package",
            "path": "contentful-resolve-response",
            "version": "1.1.4",
            "meta": {
              "originalName": "resolveContentfulResponse"
            }
          },
          "normalize": {
            "type": "package",
            "version": "git+https://github.com/ionutpasca/contentful-normaliser.git",
            "path": "@teleporthq/teleport-cms-mappers/contentful",
            "meta": {
              "namedImport": true
            }
          }
        },
        "headers": {
          "authToken": {
            "type": "env",
            "content": "CMS_ACCESS_TOKEN"
          },
          "Content-Type": {
            "type": "static",
            "content": "application/json"
          }
        },
        "path": {
          "baseUrl": {
            "type": "env",
            "content": "CMS_URL"
          },
          "route": {
            "type": "static",
            "content": "entries"
          }
        },
        "params": {
          "content_type": {
            "type": "static",
            "content": "author"
          },
          "limit": {
            "type": "static",
            "content": 10
          }
        }
      },
```

PR #789, opened 2023-05-11
(fix): Handlig names for states in cms nodes

JayaKrishnaNamburu:
We never really enforced the naming conventions for different areas in the code generation. Like
`prop` and `state` when initialised should always go with `camelCase` starting with small letters etc. This helps in two way
- consistent naming conventions
- Irrespective of whatever the prop name bugs, or any illegal characters like `-` in then. These get removed from the names.
Tested for both, strapi and contentful. So, would like to have a second check as well. And we can merge this. Parallelly needs testing in other frameworks too in-order to make sure. None of the existing things are broken. Which i am already doing 👍


JayaKrishnaNamburu:
- [x] The resolver should run on the `cms-nodes` and make sure all the dynamic connections there are also changed to follow the naming conventions

examples/uidl-samples/cms-project.json
```diff
@@ -464,45 +464,15 @@
...
-                },
-                "AuthorEntityLoading": {
-                  "id": "92fc9f25-c10b-4703-b459-f70a953719f6",
-                  "defaultValue": false,
-                  "type": "boolean"
-                },
-                "AuthorEntityError": {
-                  "id": "6f2340bf-b9dd-4ae4-8aa9-0f4eb5fdeb35",
-                  "defaultValue": false,
-                  "type": "boolean"
```
JayaKrishnaNamburu, line 505: The nodes already define a `loading`, `error` nodes to show them. So, we don't need to explicitly set them on the UIDL. Generators can populate these state or prop states and inject them in the `state` or `prop` definitions.

examples/uidl-samples/cms-project.json
```diff
@@ -1796,8 +1764,6 @@
                           }
                         ],
                         "statePersistanceName": "AuthorList1",
-                        "loadingStatePersistanceName": "AuthorListLoading1",
-                        "errorStatePersistanceName": "AuthorListError1",
                         "resourceId": "a229b188-cf3d-4393-b9c1-89249461d58b"
```
JayaKrishnaNamburu, line 1767: As above, we don't need them

vladgrecu (approved):

PR #796, opened 2023-05-28
fix/THQ-2211 - allow multiple styles in background styles 

vylly:
Split static styles and handle them separately, then concatenate them to handle them correctly

### DIFF
--- a/packages/teleport-uidl-resolver/__tests__/utils.ts
+++ b/packages/teleport-uidl-resolver/__tests__/utils.ts
@@ -17,8 +17,16 @@ import {
   checkForDefaultPropsContainingAssets,
   checkForDefaultStateValueContainingAssets,
   resolveElement,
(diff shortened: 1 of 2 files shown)


JayaKrishnaNamburu:
@vylly can you share a example for the sue-case. i don't remember doing this before. And it parser is needed for css-values we can sue `css-tree` for it. Instead of writing a custom parser.

vylly:
> 

We can have some style nodes that look like this 
`linear-gradient(90deg, rgb(189, 195, 199) 0.00%,rgba(44, 62, 80, 0.7) 100.00%)', url("/favicon-32x32-1200w.png")`
This can happen when we set a container with both a background image and a gradient background

In this case the generator is not trying to split the style node into multiple chunks but is handling only the first element which generates some issues

JayaKrishnaNamburu (approved):

PR #805, opened 2023-06-21
Allow page attribute to be taken from query params

ionutpasca:


vladnicula (approved):

JayaKrishnaNamburu:
This looks good, but IMO can we change the way we handle this. Instead of doing `paginationQueryParams` we can make a new `ReferenceType` called `router`. And then on any node you can just simply add
```json
"limit" : {
  "type": "dynamic",
  "content": {
     "referenceType": "router",
     "id": "pageId"
  }
}
```

When the node is parsed we can just add the `userRouter` dependency into the component. Just like how we add `useState` and useEffect` using state and propDefinitions. Helps in going platform agnostic, and keep in sync with all kind of references.

JayaKrishnaNamburu (commented):
Just the above `refereneType` change. All looks good 👍

ionutpasca:
Had a disscussion with @JayaKrishnaNamburu  and the implementation is going to stay as it is now

JayaKrishnaNamburu (approved):

PR #809, opened 2023-06-28
Add support for date-time nodes

ionutpasca:


ionutpasca:
We need to publish the component we are using as DateTime to npm and we're gonna change the url dependencies after

packages/teleport-component-generator-html/src/plain-html-mapping.ts
```diff
@@ -23,6 +23,19 @@ export const PlainHTMLMapping: Mapping = {
...
+    'date-time-node': {
+      elementType: 'date-time-primitive',
+      dependency: {
+        type: 'package',
+        version: '1.0.0',
+        path: 'date-time-primitive',
+        meta: {
+          importJustPath: true,
+          importAlias:
+            'https://github.com/teleporthq/date-time-primitive/blob/main/dist/default/lib.umd.js',
```
JayaKrishnaNamburu, line 35: We should use this version of the build
https://github.com/teleporthq/date-time-primitive/blob/main/dist/react/lib.mjs
UMD bundles react with IT and not proper shakeable 👍

packages/teleport-component-generator-html/src/plain-html-mapping.ts
```diff
@@ -23,6 +23,19 @@ export const PlainHTMLMapping: Mapping = {
         },
       },
     },
+    'date-time-node': {
+      elementType: 'date-time-primitive',
+      dependency: {
+        type: 'package',
+        version: '1.0.0',
+        path: 'date-time-primitive',
```
JayaKrishnaNamburu, line 31: Just dropping by, when we publish the package to `npm` this should be changed to
`@teleporthq/date-time-primitive`

vladnicula:
Shall we make a `teleport-react-component` package that contains a set of react components? Similar to how the CMS mappers repo exists? We don't need separate npm packages, just a suite a components.

JayaKrishnaNamburu (commented):
Looks good, let's publish the package and get this closed 👍. Looks like the `cms-project.json` is deleted by accident

vladnicula (approved):

JayaKrishnaNamburu:
That's actually good, we can have all frameworks. Not just react itself. We can bring `dangerous-html` and `date-time` and `lottie-node` components into one roof 👍

vladnicula:
I was thinking we could have a `react` package, with a single build system for react, a `vue` one for vue etc. The same way we have component systems like mui for react, and a separate one for vue.

PR #810, opened 2023-07-03
Use teleport packages for code-embed and date-time primitive

ionutpasca:


packages/teleport-component-generator-html/src/plain-html-mapping.ts
```diff
@@ -32,7 +33,7 @@ export const PlainHTMLMapping: Mapping = {
         meta: {
           importJustPath: true,
           importAlias:
-            'https://github.com/teleporthq/date-time-primitive/blob/main/dist/default/lib.umd.js',
+            'https://unpkg.com/@teleporthq/thq-lit-components@1.0.0/dist/thq-lit-components.umd.js',
```
JayaKrishnaNamburu, line 36: @ionutpasca we can use single repo to maintain all 3 frameworks. As vite can build all three at the same time. As the changes always need to be in sync with all the three flavours. That makes it less confusing while maintaining too.

packages/teleport-component-generator-html/src/plain-html-mapping.ts
```diff
@@ -16,10 +16,11 @@ export const PlainHTMLMapping: Mapping = {
       dependency: {
         type: 'package',
         path: 'dangerous-html',
-        version: '0.1.12',
+        version: '1.0.0',
         meta: {
           importJustPath: true,
-          importAlias: 'https://unpkg.com/dangerous-html@0.1.12/dist/default/lib.umd.js',
+          importAlias:
+            'https://unpkg.com/@teleporthq/thq-lit-components@1.0.0/dist/thq-lit-components.umd.js',
```
JayaKrishnaNamburu, line 23: And we need to improve this, or else if we are using just `dangerous-html`. It installs the `date-time-primitive` too even if we are not using. Since these are `umd` builds. The bundlers can't tree shake them

packages/teleport-component-generator-html/src/plain-html-mapping.ts
JayaKrishnaNamburu: ok, I spoke with vladut. We can keep them seperately. But, let's map these URL's to independent umd builds. Instead of loading all of them 👍

JayaKrishnaNamburu:
I will take care of this @ionutpasca. I would like to change the build system a little for the `lit` components. So they can be loaded the components on demand. Instead of all of them 👍

PR #817, opened 2023-07-18
(feat): Add support for ISR and on-demand cache validation in projects

JayaKrishnaNamburu:
This PR introduces a new field `cache` under the `uidl.resources`. It will help in handling cache mechanism for the entire project in NextJS. 

If the `resource.cache.revalidate` is set of a number. It is used in the `getStaticProps`. If the `resource.cache.revalidate` is set to a `UIDLDependency`. Then we won't be adding `revalidate` in `getStaticProps` instead we make the project to use on demand ISR.

And a new `revalidate` end-point is added to the project with the mapper that is passed with the revalidate.


vladgrecu (approved):

packages/teleport-project-generator-next/src/project-template.ts
```diff
@@ -20,9 +20,9 @@ const projectTemplate: GeneratedFolder = {
   "author": "TeleportHQ",
   "license": "MIT",
   "dependencies": {
-    "next": "^12.1.0",
-    "react": "^17.0.2",
-    "react-dom": "^17.0.2"
+    "next": "^13.4.10",
```
ionutpasca: Are we sure we support this version?

packages/teleport-test/src/standalone.ts
```diff
@@ -91,6 +92,16 @@ const run = async () => {
           ...packerOptions.publishOptions,
           projectSlug: 'teleport-project-next-cms',
         },
+        plugins: [
+          new ProjectPluginCustomFiles([
+            {
+              id: 'teleport-config.json',
+              name: 'teleport-config.json',
+              path: [],
+              content: JSON.stringify({ '/about': { path: '/about' } }, null, 2),
```
ionutpasca: Why do we need this? I'm just trying to understand what happens under the hood...

packages/teleport-types/src/generators.ts
```diff
@@ -139,7 +139,9 @@ export interface GeneratorOptions {
   designLanguage?: {
     tokens?: UIDLDesignTokens
   }
-  resources?: { items: UIDLResources['items'] } & { path: ProjectStrategy['resources']['path'] }
+  resources?: { items: UIDLResources['items']; cache: UIDLResources['cache'] } & {
```
ionutpasca, line 142: So....each resource will have it's own cache validator? Do I understand correctly?

packages/teleport-test/src/standalone.ts
JayaKrishnaNamburu: So, the mapping between `contentType` from the CMS and the `route` in the projects are mutable. Means, users can change them any time.
Eg: contentType of `BlogPost` can become `/some-random-route`. 

This info, is only known to our backend and not contentful. So, contentful webhook is always going to call with `BlogPost`. In order to figure these mappings out. GUI is going to inject a file called `teleport-config.json`. Which is picked up by the `webhook` handler and then maps the contentTypes to their respective routes. More about the implementation is here
https://github.com/teleporthq/teleport-cms-mappers/pull/3/files#diff-8c76daa4bb77de60c6b376dfdbbf7faec4446fbcce1c5baf9169dcc7aae32561

packages/teleport-types/src/generators.ts
JayaKrishnaNamburu: They need to be defined in the project level for now. This is just to pass to all the `component-plugins` by taking from the `global`. So, each component can use that value to set the timeout in `revalidate`

packages/teleport-project-generator-next/src/project-template.ts
JayaKrishnaNamburu: Yes, the current projects works as expected. And till now, we deployed only static sites. So, it's not going to break any projects. We need `next@13` in order to clear cache for multiple routes. `next@12` allows us only to clear a single route. 
But we generate `listing`, `details` pages for each content type. And we need to clear cache for the both, or else one of it will keep showing the data from the build time.

packages/teleport-test/src/standalone.ts
JayaKrishnaNamburu: This particular change here is just to test the feature out. But in general this is exactly the place where GUI injects it's file. As a fallback just in case if GUI fails to inejct but the `revalidate` is enabled. We inject a empty JSON file
https://github.com/teleporthq/teleport-code-generators/pull/817/files#diff-f81aa656e26fb0ab234fcf65a5c0624f75b35a9430b1e32684f570f91a955b87R43

Reason, we use static imports
```js
import routeMapper from '../../teleport-config.json'
```

Which breaks the whole build because of NextJS static analysis at build time if the file is missing. We can't use `readFileSync` because the file is not added to the `.next` at the end of build process. Webpack includes a file only if is `imported` so. We need to always ensure there is some sort of file if not the injected file to make the builds pass.

Probably not needed the whole explanation 😅, but just wrote it anyways. As we can use as future reference 😀

packages/teleport-project-generator-next/src/project-template.ts
JayaKrishnaNamburu: I am going to downgrade again, found a workaround in next@12 itself. 13 has a lot of bugs with cache 👍

packages/teleport-project-generator/src/index.ts
```diff
@@ -332,6 +332,7 @@ export class ProjectGenerator implements ProjectGeneratorType {
           this.strategy?.resources?.path && {
             resources: {
               items: uidl?.resources?.items,
+              cache: uidl.resources.cache,
```
vladnicula: if `uidl?.resources?.items,` needs `?` then `uidl.resources.cache` needs `?` or none of them do.

packages/teleport-project-generator/src/index.ts
```diff
@@ -634,6 +635,7 @@ export class ProjectGenerator implements ProjectGeneratorType {
       collectedDevDependencies = { ...collectedDevDependencies, ...runAfterResult.devDependencies }
       inMemoryFilesMap = runAfterResult.files
     } catch (e) {
+      console.trace(e)
```
vladnicula, line 638: we need this in the merge?

packages/teleport-project-plugin-next-cache/tsconfig.json
vladnicula, line 1: the naming `next-cache` is unclear. What cache is this? Are we caching some kind of information in memory and returning it instead of computing it again or fetching again? If not, we should use a different name. 

since this plugin creates a revalidation api endpoint, I think we should name it something related to revalidation, and api endpoints.

packages/teleport-test/src/standalone.ts
```diff
@@ -91,6 +93,19 @@ const run = async () => {
           ...packerOptions.publishOptions,
           projectSlug: 'teleport-project-next-cms',
         },
+        plugins: [
+          new ProjectPluginNextCache({
+            routeMappers: {
+              /* tslint:disable no-invalid-template-strings */
+              bogpost: ['/bogpost/${id}', '/bogpost'],
```
vladnicula, line 100: I assume this is for testing only, and in production we'll get this data from GUI, right?

packages/teleport-types/src/generators.ts
vladnicula: @ionutpasca is this response clear? please resolve again if you agree.

packages/teleport-test/src/standalone.ts
JayaKrishnaNamburu: yes 👍

packages/teleport-project-generator/src/index.ts
JayaKrishnaNamburu: Yes, right now. It is encapsulating the actual origin of error. We do have this already for the `runBefore` errors. Just added for `runAfter` too

packages/teleport-project-generator/src/index.ts
JayaKrishnaNamburu: We don't have `default` values for `items` So, the usual projects might not even have any resources in them. But we do have `default` values for `cache`. If there is no cache in the UIDL. By default the value is `60` so typescript is all happy anytime we read `uid.resources.cache`

vladnicula: I was referring to the code, not application knowledge. If `uidl?` on first line, then `uidl?` on second line as well.

packages/teleport-project-plugin-next-cache/tsconfig.json
JayaKrishnaNamburu: suggest one then, how about `@teleporthq/teleport-next-revalidate-api` ?

vladnicula: sounds good!

PR #818, opened 2023-07-18
Handle CMS list repeater separately from CMS list node

TudorCe:
- Add support for a new UIDL primitive -> cms-list-repeater, which maps to the Repeater element from teleporthq-react components package
- Change CMS List node implementation to no longer generate the Repeater element itself. It is now more similar to the CMS Item node, and lets the success slot decide what it generates for itself.
- Enable `expr` type conditional nodes
- Change a bit the `getStaticProps` generation for CMS Listing pages to now accommodate new changes brought to the normalize function
- Add extra parameter to normalize function

### DIFF
--- a/packages/teleport-project-generator/src/resource.ts
+++ b/packages/teleport-project-generator/src/resource.ts
@@ -20,17 +20,8 @@ export const resourceGenerator = (
   let returnStatement: types.Identifier | types.CallExpression = types.identifier('response')
 
   resource.mappers.forEach((mapper) => {
-    // TODO: is this fine here? Sure doesn't look like production-worthy code, but I'm not 100% sure how else to tackle this atm.
-    // Essentailly, if it's a normalize function for the CMS, it now expects two parameters instead of one.
-    if (mapper === 'normalize') {
-      const extraParamsStatement = types.identifier('params')
-      returnStatement = types.callExpression(types.identifier(mapper), [
-        returnStatement,
-        extraParamsStatement,
-      ])
-    } else {
-      returnStatement = types.callExpression(types.identifier(mapper), [returnStatement])
-    }
+    // Fallback value for returnStatement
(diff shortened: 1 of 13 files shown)


JayaKrishnaNamburu:
Can you share some more details @TudorCe regarding the change and expected behaviour etc

JayaKrishnaNamburu:
@TudorCe is the `cms-list-repeator` is what responsible for generating pagination buttons ?

TudorCe:
> 

Nono, the pagination buttons currently have nothing custom in code generators, they are generated and passed here as regular links from the GUI

packages/teleport-project-generator/src/resource.ts
```diff
@@ -20,7 +20,17 @@ export const resourceGenerator = (
...
+    // Essentailly, if it's a normalize function for the CMS, it now expects two parameters instead of one.
+    if (mapper === 'normalize') {
+      const extraParamsStatement = types.identifier('params')
+      returnStatement = types.callExpression(types.identifier(mapper), [
+        returnStatement,
+        extraParamsStatement,
+      ])
+    } else {
+      returnStatement = types.callExpression(types.identifier(mapper), [returnStatement])
+    }
```
JayaKrishnaNamburu: Regarding this one, looks like the `params` are not being sent all the time. We need to add a new value to UIDL. Which says the `params` additional to the actual content that can be taken. 
https://github.com/teleporthq/teleport-code-generators/blob/feat/better-fetch/packages/teleport-types/src/uidl.ts#L67C36-L67C50

As even now, in few cases the params are being sent and few places no. So,

```js
normalize(resolveContentfulResponse(response))
```

```js
normalize(response, params)
```

Let's discuss, we should change this 👍

JayaKrishnaNamburu (changes requested):

JayaKrishnaNamburu (approved):

PR #821, opened 2023-07-24
Return only the actual data for cms items, without pagination meta

ionutpasca:


packages/teleport-project-plugin-inline-fetch/src/next/utils.ts
```diff
@@ -211,7 +211,11 @@ const computeUseEffectAST = (params: {
             'data',
             ...itemValuePath,
           ]) as types.MemberExpression)
-        : types.memberExpression(types.identifier('data'), types.numericLiteral(0), true)
+        : types.memberExpression(
+            types.memberExpression(types.identifier('data'), types.identifier('data'), false),
```
JayaKrishnaNamburu, line 215: @ionutpasca test all the CMS's once. I remember having a little issue with this one. Are the mappers going to change as well ? Then it's fine.

JayaKrishnaNamburu: We are using `meta` as well. We changed this in the recent `data` and `meta` feature

ionutpasca: I'm not sure what changes are done inside the mappers....with the version of the mappers that is currently used in the generators, the problem that we're having is fixed. If what you're referring to "meta" are the pagination fields, then this is not a problem...because you don't have any pagination meta inside a cms-item....only in cms-list.

Does it make sense?

JayaKrishnaNamburu (approved):

PR #823, opened 2023-07-27
(feat): Add support for external resource fetchers

elizamaria:
Add support for external resource fetchers inside getStaticPaths and getStaticProps
Add support for importDefinitions on the 'conditional' node, which will get imported in the page. Support for importDefnitions on component nodes already exists
Fix a bug when multiple imports are from the same source. 
Not necessary for now, but also fixed support for custom headers


packages/teleport-plugin-common/src/utils/ast-utils.ts
```diff
@@ -610,19 +610,33 @@ export const generateRemoteResourceASTs = (resource: UIDLResourceItem) => {
...
+    allHeaders.push(authHeaderAST)
+  }
+
+  if (headersASTs.length) {
+    allHeaders = allHeaders.concat(headersASTs)
+  }
+
+  const headers = types.objectProperty(
+    types.identifier('headers'),
+    types.objectExpression(allHeaders.length ? [...allHeaders] : [])
```
JayaKrishnaNamburu: You can directly use it this way

```js
types.objectExpression(allHeaders)
```

as it is being declared as empty array anyways

packages/teleport-plugin-next-static-props/src/index.ts
```diff
@@ -19,25 +18,49 @@ export const createStaticPropsPlugin: ComponentPluginFactory<StaticPropsPluginCo
...
+
+      const importPath = `${resources.path}${StringUtils.camelCaseToDashCase(usedResource.name)}`
+
+      dependencies[resourceImportName] = {
+        path: importPath,
+        /*
+          Local resources can be marked as a library. As they will be ignored in adding to package.json at the end.
+        */
+        type: 'library',
+        version: 'latest',
```
JayaKrishnaNamburu: I just realised, we can mark the type as `local` for local dependency. So, we won't accidentally mark local imports as deps.

```js
{
  path: '../.../<function>',
  type: 'loca'
}
```

And for external like caisey or others we can mark as

```js
{
  type: 'package',
  version: '..',
  path: '<absolute-path>'
}
```

JayaKrishnaNamburu (approved):
Just minor 1-2 changes, Looks Good 🚀

packages/teleport-plugin-next-static-props/src/index.ts
elizamaria: You mean to replace library with local?

JayaKrishnaNamburu: Yup 👍

PR #824, opened 2023-07-28
Re-add pagination meta to static props

TudorCe:
- Adds back a few lines of code that were removed in the latest merge due to conflicts


JayaKrishnaNamburu (approved):
Good catch 💯

PR #826, opened 2023-08-01
Fix and update CMS node parsing to not have ifs in a switch

TudorCe:
- Updates parseComponentNode function to also parse CMS Node item's slots, which fixes generation issues with their direct children
- Prettify code a bit by moving each cms node in its own case (except list and item which are identical)


JayaKrishnaNamburu (approved):
Looks good 🚀

PR #840, opened 2023-08-21
(feat): Extract root resources from the the cms-list and cms-item

JayaKrishnaNamburu:
All the resources from that are in the root level in a component and doesn't use any dynamic values will be extracted.
The extraction will not happen if
- The params passed to the node is using `expressions` to send them.
- Once the #841 is merged, we should omit the `state` / `prop` is being used in the resource params.
- Currently the extraction is also omitted for external resource fetchers. Which means all `caisy` projects will be un-effected.
- Resources that are extracted will not have the wrappers created inside `pages/api` folder.


examples/uidl-samples/cms-project.json
```diff
@@ -2555,121 +1848,257 @@
...
+                                      }
+                                    ]
+                                  }
+                                }
+                              ]
+                            }
+                          }
+                        },
+                        "valuePath": [],
+                        "itemValuePath": ["data"],
```
JayaKrishnaNamburu: We are not sending these values for the `cms-item` and `cms-list`. In Listing and details page. They are being handled using `exposeAs` attribute. So, this need to be corrected in the UIDL export from the platform

PR #842, opened 2023-08-22
Use id as data provider key for main cms items

ionutpasca:


packages/teleport-plugin-common/src/node-handlers/node-to-jsx/index.ts
```diff
@@ -332,10 +332,20 @@ const generateCMSNode: NodeToJSX<UIDLCMSListNode | UIDLCMSItemNode, types.JSXEle
       )
     )
 
+    let keyValue = 'props?.pagination?.page'
+
+    if (node.type === 'cms-item') {
+      const { entityKeyProperty } = node.content
+      const entityName = initialData.content.id
```
JayaKrishnaNamburu, line 339: There is a catch here 🤔, `cms-list` and `cms-item` don't have `initialData` all the time. Maybe the `renderPropItenditifer` or even `index` is better option ?

elizamaria: @JayaKrishnaNamburu after this is merged, let's do a version update for the better-fetch as we need the last merge to be pushed in prod for meta tags support

ionutpasca: @JayaKrishnaNamburu , indeed, only cms-item has initialData. There is a problem there, as there is a 'key' property placed on the `DataProvider` component in the generated code for CMS detail pages main nodes. That key was always set to 'props?.pagination?.page', which is not correct. 

If we are talking about a cms main node for a details page, we need to use a property from that CMS record that is fetched from the provider. We cannot use a hardcoded 'props?.ENTITY_NAME?.id' value, as we cannot be sure if there is actually an 'id'. So we need to send from the UIDL, what differentiator that details page has. It might as well be the 'slug', or some other property that the user chooses.

Let me know if this is clear or not...

JayaKrishnaNamburu: @ionutpasca i mean, the `entityName` is being read from `initialData.content.id`. In case of `DataProvider` loading using `fetcher`. We don't have the `initialData.content.id` right. Do we have a fallback for those cases ?

ionutpasca: Not sure I understand what you're saying @JayaKrishnaNamburu . We do have `initialData.content.id`. This is the name of the prop that is sent from getStaticProps that contains the actual CMS entity data. What is inside `entityKeyProperty` is the name of the property from  that CMS entity that is used for creating the details pages unique.

Screenshot 2023-08-23 at 14 16 22

What you see in this image is the `[id]`. If you take this example, `entityKeyProperty` will equal to `id`. If the details pages are generated using a different property, let's say `slug`, then `entityKeyProperty === 'slug'`. 

Does it make sense?

PR #848, opened 2023-08-30
(chore): Build all cms-projects in the CI

### LINKED ISSUE #844: Use GitHub actions to build CMS projects
Right now, managing and making sure all CMS projects build has become really tricky. We need to run some scripts that can build the projects and build them. To make sure the if the UIDL remains the same, the projects build all the time.
For inspiration, here is a attempt from me previously
https://github.com/JayaKrishnaNamburu/teleport-code-generators/pull/3

JayaKrishnaNamburu:
fixes #844


packages/teleport-project-plugin-inline-fetch/src/next/utils.ts
```diff
@@ -215,11 +215,24 @@ export default async function handler(req, res) {
         const functionDecleration = (getStaticPropsChunk.content as types.ExportNamedDeclaration)
           .declaration as types.FunctionDeclaration
         const functionBody = functionDecleration.body.body
-        functionBody.unshift(...declerations)
+        const tryBlock = functionBody.find(
```
JayaKrishnaNamburu: Ignore all the changes of this file. They are already handled in #849

PR #851, opened 2023-09-07
(refactor): Wrap `getStaticPaths` with try-catch to handle un-expected errors from CMS

JayaKrishnaNamburu:
The PR wraps the `getStaticPaths` API fetches with `try-catch`. So, any un-expected behaviours from the responses will not break the build.

It happens in wordpres, it responds with `code:401` insead of throwing error for the API call. And that is breaking the build for rest of the pages. This will just disable the `pre-rendering` for the pages that have issues. And still makes the project deployable and executable at runtime.

```js
export async function getStaticPaths() {
  try {
    const response = await navMenuItemPageInitialPathsA0fe0Resource({})
    return {
      paths: (response || []).map((item) => {
        return {
          params: {
            id: (item?.id).toString(),
          },
        }
      }),
      fallback: 'blocking',
    }
  } catch (error) {
    return {
      paths: [],
      fallback: 'blocking',
    }
  }
}
```

### DIFF
--- a/examples/uidl-samples/wordpress.json
+++ b/examples/uidl-samples/wordpress.json
@@ -1,8 +1,8 @@
 {
-  "name": "wordpress",
+  "name": "Rotating Radiant Ape",
   "globals": {
     "settings": {
-      "title": "wordpress",
+      "title": "Rotating Radiant Ape",
       "language": "en"
     },
     "assets": [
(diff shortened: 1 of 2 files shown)


JayaKrishnaNamburu:
I updated the `wordpress.json` file too. This is the UIDL that breaks the deployments with the specific error. And since the CI can run and do a prod build for these projects. We can catch if the CI is failing too 👍

vladnicula:
You could simplify that code a bit

```tsx
export async function getStaticPaths() {
  let response = []
  try {
    response = await navMenuItemPageInitialPathsA0fe0Resource({})
  } catch (error) {
    console.error('static paths something something dark side')
  }

  return {
      paths: (response || []).map((item) => {
        return {
          params: {
            id: (item?.id).toString(),
          },
        }
      }),
      fallback: 'blocking',
    }
}
```

JayaKrishnaNamburu:
The thing is, the API from wordpress is respond with `200` and inside body with `status: { code: 401 }`. So the fetch doesn't actually throw error. Instead it happens in `paths: (response || []).map((item)` calculation.

JayaKrishnaNamburu:
Same with the `getStaticProps`, so started wrapping the whole with try-catch there to, reference #846.

vladnicula:
I would like to have a log in the catch so we know what broke the build instead of a silent failure.

I would also like to see fallback block only once so I don't have to ask myself if we are doing more stuff than just catching the error for map.

JayaKrishnaNamburu:
> I would like to have a log in the catch so we know what broke the build instead of a silent failure.
> 
> I would also like to see fallback block only once so I don't have to ask myself if we are doing more stuff than just catching the error for map.

Yeah sounds good, i will refactor both `getStaticPaths` and `getStaticProps` to log these errors and the syntax too 👍

JayaKrishnaNamburu:
Updated the syntax now @vladnicula. It looks like this now

```js
export async function getStaticPaths() {
  const response = await navMenuItemPageInitialPathsA0fe0Resource({})
  let paths = []
  try {
    paths = (response || []).map((item) => {
      return {
        params: {
          id: (item?.id).toString(),
        },
      }
    })
  } catch (error) {
    console.log('Pre-rendering failed for the route')
    console.log(error)
  }
  return {
    paths: paths,
    fallback: 'blocking',
  }
}
```

We can just skip the page that is crashing, since 95% of the times. It something that is working in GUI it definitely works in deployment too. Just some edge cases in wordpress. It might be too much to break the full build. If a specific path fails. What say ?

vladnicula:
looks good, merging

PR #852, opened 2023-09-08
Add source support for repeater node

ionutpasca:


JayaKrishnaNamburu:
@ionutpasca can you add a example in the PR description please. Because any source for the `Repeator` node will effect the return of `DataProvider` too right.

ionutpasca:
@JayaKrishnaNamburu , the **Repeater** will not be affetcted at all by this MR. What I did was only to add support for using the **Repeater** component without the need of a **DataProvider**. Before, the Repeater was only used inside a DataProvider, which exposed a variable named **params**. The Repeater would take this `params` and it would display an element for each element from it.

With this change, we have not the possibility to use the Repeater with no dependency from the DataProvider.
Let's say you have an array named `foo`. We can say now that we need to use the Repeater and loop through `foo`, not `params`, items (`foo` being the source).

Does it make sense?

JayaKrishnaNamburu:
Yup, sounds good 👍

JayaKrishnaNamburu (approved):

JayaKrishnaNamburu:
Maybe we can mark the `node.content.source` as `UIDLStaticNode` or `UIDLDynamicNode` just to keep the consistency. We don't directly using strings in UIDL. Instead `static` / `dynamic`. Not a deal breaker, but keeps the consistency 😄

ionutpasca:
@JayaKrishnaNamburu what do you mean `we can't directly use strings in UIDL`. In the same interface, we have `  renderPropIdentifier` which is a string. I see this source the same as we are using `renderPropIdentifier`. I can change it into a static node, but it won't keep the consistency any more than it is doing now....Does it make sense or not?

JayaKrishnaNamburu:
> @JayaKrishnaNamburu what do you mean `we can't directly use strings in UIDL`. In the same interface, we have ` renderPropIdentifier` which is a string. I see this source the same as we are using `renderPropIdentifier`. I can change it into a static node, but it won't keep the consistency any more than it is doing now....Does it make sense or not?

It does, `renderProp` should have been `static` node too. But, render prop can always be just a string. Here we can connect to resource to a `prop` or a `state` in future if needed right. So, thought we can keep an option. But, the change is always backward compatible. So, it's fine as it is now 👍

ionutpasca:
@JayaKrishnaNamburu , let's merge it as it is now and if we're going to find the need to change it and support both dynamic and static values, we can improve it. What do you say?

JayaKrishnaNamburu:
I was thinking, you will do the merge and just leaving comments 😅 after approving. Going to merge now 👍

PR #854, opened 2023-09-18
(feat): Allow exporting components as pages with a flag in html-project-generators

JayaKrishnaNamburu:
The default behaviour will keep exporting components. When passing a flag, the components also inherit the global into the components and works independently.


TudorCe (approved):
Looks good 👍

PR #856, opened 2023-09-20
(fix): Use Redirect and Switch to define fallback route in react

JayaKrishnaNamburu:
The fallback route works only when `Switch` and `Redirect` are used in react. So, this fix now generates the routes to 
```jsx
const App = () => {
  return (
    <Router>
      <Switch>
        <Route component={Home} exact path="/" />
        <Route component={About} exact path="/about" />
        <Route component={Fallback} path="**" />
        <Redirect to="**" />
      </Switch>
    </Router>
  )
}
```

### DIFF
--- a/packages/teleport-test/src/standalone.ts
+++ b/packages/teleport-test/src/standalone.ts
@@ -1,4 +1,4 @@
-import { readFileSync } from 'fs'
+import { readFileSync, rmdirSync, mkdirSync } from 'fs'
 import { join } from 'path'
 import chalk from 'chalk'
 import { packProject } from '@teleporthq/teleport-code-generator'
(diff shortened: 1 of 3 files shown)


packages/teleport-test/src/standalone.ts
```diff
@@ -72,8 +72,8 @@ const log = async (cb: () => Promise<string>) => {
 const run = async () => {
   try {
     if (packerOptions.publisher === PublisherType.DISK) {
-      rmdirSync('dist', { recursive: true })
-      mkdirSync('dist')
+      // rmdirSync('dist', { recursive: true })
```
TudorCe: I think you can revert this :D

JayaKrishnaNamburu: Ah, doing it 👍

JayaKrishnaNamburu: Done 👍

PR #861, opened 2023-09-26
(fix): Extract styles on the node while parsing the html-node element

JayaKrishnaNamburu:
We wrap the code-emebeds with `div` in the canvas. And then even allow users to apply styles on them. But the problem was, when we are extracting them to UIDL. We are passing them as

```json
{
  "type": "element",
  "content": {
    "elementType": "html-node",
    "referencedStyles": {},
    "attrs": {
      "html": {
        "type": "raw",
        "content": "<div>\n    This is a simple div element\n</div>"
      }
    },
    "style": {
      "backgroundColor": {
        "type": "dynamic",
        "content": {
          "referenceType": "token",
          "id": "--dl-color-success-300"
        }
      }
    },
    "children": []
  }
}
```

which is expected. As the styles lie next to the node. But we are loosing the hierarchy that comes by when the rendering is happening in iframe. So, we need to extract the styles in the same way and try to generate similar hierarchy.
Screenshot 2023-09-26 at 22 20 18


examples/test-samples/project-invalid-sample.json
```diff
@@ -314,4 +314,4 @@
       }
     }
   }
-}
\ No newline at end of file
+}
```
JayaKrishnaNamburu, line 317: We have `prettier` up and running now, before we try to commit anything. That's the reason for all these fixes. We used to have `pretty-quick` but the `prettier` needed for `pretty-quick` to run was missing before.

PR #867, opened 2023-10-09
(feat): Extract root resources for data fetching from external resources

JayaKrishnaNamburu:
Till now, we extracted the root resources only for inline fetches. But for integrations which are using external libraries are not. Because we don't clearly know what `process` variables the external functions are using etc. But now, are opting them to extract as well. As currently there is only one integration that is using it. And we know we are not using any external stuff. This improves caisy integration 👍

### DIFF
--- a/packages/teleport-project-plugin-inline-fetch/src/next/utils.ts
+++ b/packages/teleport-project-plugin-inline-fetch/src/next/utils.ts
@@ -74,14 +74,13 @@ export const createNextComponentInlineFetchPlugin: ComponentPluginFactory<Contex
         let dependencyPath: string
 
         if ('id' in extractedResource) {
-          resourceName = StringUtils.createStateOrPropStoringValue(
-            resources.items[extractedResource.id].name + 'Resource'
-          )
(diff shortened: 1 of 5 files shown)


packages/teleport-project-plugin-inline-fetch/src/next/utils.ts
```diff
@@ -74,19 +69,29 @@ export const createNextComponentInlineFetchPlugin: ComponentPluginFactory<Contex
...
-        const resourcesPath = paths.resources
-        const currentPagePath = [...paths.pages, ...uidl.outputOptions.folderPath]
+        const extractedResource = extractedResources[propKey]
+        let resourceName: string
+        let dependencyPath: string
 
-        dependencies[resourceName] = {
-          type: 'local',
-          path:
+        if ('id' in extractedResource) {
```
JayaKrishnaNamburu, line 76: This is to make `TS` happy, the regular `extractedResource?.id` doesn't seem to detect the resource as a local resource.

vladnicula (approved):

PR #877, opened 2023-12-11
(fix): Missing project style imports when using project styles in css-in-js flavours and fix static-classes on node.

JayaKrishnaNamburu:
When generating projects with any `css-in-js` style of frameworks. The import of the global style sheet file is broken right now.

```js
import projectStyles from 'pagesstyle.module.css'
import styles from './about.module.css'
```

See the `pagesstyle.module.css` should be `./style.module.css`. Instead of calculating the relative path from the component/page w.r.t to the global style sheet location. We are just appending them. This fixes the code to generate

```js
import projectStyles from './style.module.css'
import styles from './about.module.css'
```

### DIFF
--- a/packages/teleport-plugin-common/src/utils/ast-utils.ts
+++ b/packages/teleport-plugin-common/src/utils/ast-utils.ts
@@ -106,17 +106,21 @@ export const addDynamicAttributeToJSXTag = (
 export const addMultipleDynamicAttributesToJSXTag = (
   jsxASTNode: types.JSXElement,
   name: string,
-  attrValues: Array<types.MemberExpression | types.Identifier> = [],
+  attrValues: Array<types.MemberExpression | types.Identifier | types.StringLiteral> = [],
   t = types
 ) => {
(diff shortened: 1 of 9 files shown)


JayaKrishnaNamburu:
This PR also fixes a bug, where users just want to add a sample `class` on the node. But they get prefixed by the css-in-js specific identifiers. 

```jsx
const SimpleComponent = (props) => {
  return (
    <div>
      <div className={` ${styles['container1']} ${styles['default-container']} `}>
        A element with local and project style
      </div>
    </div>
  )
}
```

The `default-container` is just a static string that we want to add, but the `styles` get prefixed since its `css-modules`. To fix this issue. We just need to remove the framework speific prefixes. As the use-case is to just add a class name on the node and nothing else. Which results in

```jsx
const SimpleComponent = (props) => {
  return (
    <div>
      <div className={` ${styles['container1']} ${'default-container'} `}>
        A element with local and project style
      </div>
    </div>
  )
}
```

This bug is what breaking the slider primitive in our platform when exported with other style combnations other than css

PR #890, opened 2024-02-20
add support for bracket notation path in propDefinitions

guvvanch:
Fixes validator for the case where dynamic prop has a path with bracket notation. For example: previously, the path `fields['test name']` was not supported even if it was defined in propDefinitions. This changes fix that issue

### DIFF
--- a/packages/teleport-uidl-validator/src/validator/utils.ts
+++ b/packages/teleport-uidl-validator/src/validator/utils.ts
@@ -160,7 +160,12 @@ export const checkDynamicDefinitions = (input: Record<string, unknown>) => {
         ) {
           if (styleRef.content.content.content.referenceType === 'prop') {
             const referencedProp = styleRef.content.content.content.id
-            if (!dynamicPathExistsInDefinitions(referencedProp, propKeys)) {
+            if (
+              !dynamicPathExistsInDefinitions(
+                referencedProp,
+                input.propDefinitions as ComponentUIDL['propDefinitions']
+              )
+            ) {
               const errorMsg = `"${referencedProp}" is used but not defined. Please add it in propDefinitions ${input.name}`
               errors.push(errorMsg)
               return
@@ -190,7 +195,12 @@ export const checkDynamicDefinitions = (input: Record<string, unknown>) => {
     }
 
     if (node.type === 'dynamic' && node.content.referenceType === 'prop') {
-      if (!dynamicPathExistsInDefinitions(node.content.id, propKeys)) {
+      if (
+        !dynamicPathExistsInDefinitions(
+          node.content.id,
+          input.propDefinitions as ComponentUIDL['propDefinitions']
+        )
+      ) {
         const errorMsg = `\n "${node.content.id}" is used but not defined in ${
           input.name
         } component. Please add it in propDefinitions.
@@ -205,7 +215,12 @@ Used on Node ${JSON.stringify(node)}.\nParent node is ${JSON.stringify(parent, n
     }
 
     if (node.type === 'dynamic' && node.content.referenceType === 'state') {
-      if (!dynamicPathExistsInDefinitions(node.content.id, stateKeys)) {
+      if (
+        !dynamicPathExistsInDefinitions(
+          node.content.id,
+          input.stateDefinitions as ComponentUIDL['stateDefinitions']
+        )
+      ) {
         console.warn(
           `\n"${node.content.id}" is used but not defined. Please add it in stateDefinitions`
         )
@@ -218,7 +233,12 @@ Used on Node ${JSON.stringify(node)}.\nParent node is ${JSON.stringify(parent, n
     }
 
     if (node.type === 'import') {
-      if (!dynamicPathExistsInDefinitions(node.content.id, importKeys)) {
+      if (
+        !dynamicPathExistsInDefinitions(
+          node.content.id,
+          input.importDefinitions as ComponentUIDL['importDefinitions']
+        )
+      ) {
         const errorMsg = `\n"${node.content.id}" is used but not defined. Please add it in importDefinitions`
         errors.push(errorMsg)
       }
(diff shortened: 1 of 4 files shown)


packages/teleport-uidl-validator/src/validator/utils.ts
```diff
@@ -251,16 +271,40 @@ Used on Node ${JSON.stringify(node)}.\nParent node is ${JSON.stringify(parent, n
   return errors
 }
 
-const dynamicPathExistsInDefinitions = (path: string, defKeys: string[]) => {
-  if (!path.includes('.')) {
-    // prop/state is a scalar value, not a dot notation
-    return defKeys.includes(path)
-  }
+const dynamicPathExistsInDefinitions = (
```
JayaKrishnaNamburu, line 278: Hello @guvvanch thanks for working on this, can we modify the function a little bit.
To support both kinds of notions like.

```js
hero.bgImage.src

hero['bgImage']['src']
```

So, we don't support just only `[]` notion while referring to props. Something like this
```js
function isValidObjectPath(path) {
  // Extract the keys from the path string
  const keys = path
    // Replace [' and '] with '.' for easier splitting
    .replace(/\['/g, '.')
    .replace(/'\]/g, '')
    // Replace . notation with . after replacing [] notation
    .replace(/\./g, ' . ')
    // Split the keys
    .split(/\[|\]|\./)
    .filter(Boolean);

  let obj = propDefinitions;

  for (const key of keys) {
    // Check if the key exists in the current object
    if (!obj[key]) {
      return false;
    }

    // Move to the next nested object
    obj = obj[key];
  }

  return true;
}

```

And add some tests too, please

JayaKrishnaNamburu (changes requested):

packages/teleport-uidl-validator/src/validator/utils.ts
guvvanch: Hi @JayaKrishnaNamburu , thank you for reviewing this PR. Regarding your feedback:
1) 
```
  const keys = path
    // Replace [' and '] with '.' for easier splitting
    .replace(/\['/g, '.')
    .replace(/'\]/g, '')
    // Replace . notation with . after replacing [] notation
    .replace(/\./g, ' . ')
    // Split the keys
    .split(/\[|\]|\./)
    .filter(Boolean);
```
Upon review, it appears that this implementation has some limitations. For instance, it fails to accommodate paths with double quotes, such as `hero["bgImage"]["src"]`. Additionally, the replacement of dots `.` with spaces ` . ` (`.replace(/\./g, ' . ')`) might cause issues during key comparison due to the added spaces.

To address these issues, I've introduced a more comprehensive regular expression that handles various quotation marks and ensures more accurate key extraction:
 `const pathKeys = path.split(/\.|\[\s*['"]?(.+?)['"]?\s*\]/).filter(Boolean)`
 
 2)
I've revised the construction of the `obj` variable to account for the differences in interfaces between state/prop definitions and importDefinitions:
 
```
   let obj = Object.keys(definitions).reduce((acc, key) => {
    acc[key] =
      type === 'import'
        ? definitions[key]
        : (definitions[key] ).defaultValue

    return acc
  }, {} )
```

3)
```
  for (const key of keys) {
    // Check if the key exists in the current object
    if (!obj[key]) {
      return false;
    }

    // Move to the next nested object
    obj = obj[key];
  }
```

Here, I believe, it is better to use `if (!(key in obj)) {` instead of `  if (!obj[key]) {`, so that it doesn't return `false` when the given path exists, but its value is set to `''`, `undefined`, or `null`. 

4) These changes should be covered in test cases as I added these cases to sample project data.

Please let me know if you have any further suggestions for improvement.

JayaKrishnaNamburu:
@guvvanch thanks for looking into it and the explanation. Were you able to generate a project with the same in `propDefinitions` with objects and array. Looks like the actual code-generation is missed. Like, we need to individually implement the `object` and `array` behaviour in `react-base`, `angular-base` and `vue-base` respectively. Can you look into that too please

guvvanch:
@JayaKrishnaNamburu I tested the `dynamicPathExistsInDefinitions` function itself by passing mock propDefinitions with the same interface used in UIDL. 
image


For the following part of your comment, I am not sure if I understood it, would you mind elaborating a little bit?
>  Looks like the actual code-generation is missed. Like, we need to individually implement the object and array behaviour in react-base, angular-base and vue-base respectively.

JayaKrishnaNamburu:
@guvvanch thanks for the explanation. 
I am actually looking for an example. On how these object notion is used when the components are generated. 
You can add some test cases under `base-plugins` of each package. To show on how these nodes are generated at code-generation time.

Or, are you looking for just the validation itself. And not using these nodes in component generation. 
As an inspiration, if you want to implement too. You can refer to `expr` node in teleport-types.
https://github.com/teleporthq/teleport-code-generators/blob/development/packages/teleport-types/src/uidl.ts#L313

Eg: 
Test cases to test the AST
https://github.com/teleporthq/teleport-code-generators/blob/development/packages/teleport-plugin-react-base-component/__tests__/index.ts#L46

Test cases to test the code itself
https://github.com/teleporthq/teleport-code-generators/blob/development/packages/teleport-component-generator-react/__tests__/end2end/index.ts#L120

guvvanch:
@JayaKrishnaNamburu I didn't make any changes to the code generators. 

The issue before was that UIDL wasn't passing the validation because of bracket notation path values for props/states (e.g. `fields['new name']`) even though it is  a valid object path. So, I updated the validator, added the cases I am mentioning into the sample uidl used in validator test cases as well as added this case to UIDL sample that is being used in test cases for component generators (NOTE: same sample UIDL is used for all frameworks). 

If I didn't address your questions/feedback, I am happy to have a quick meeting to discuss. Thank you!

JayaKrishnaNamburu (approved):

JayaKrishnaNamburu:
@guvvanch i just tested the changes in local once again. And it seems to work fine. Thanks for working on this and helping to get through the PR 😄

PR #892, opened 2024-02-23
(feat): Enable plugin parse embeds for all frameworks when enabled

JayaKrishnaNamburu:
- Enabled the `parse-embed` plugin for all the frameworks now. And deleted the `plugin-external-embeds` plugin too. As we are not using it anymore now.
- Fixed a bug with `styled-components`. When we have a node with dependency, we need the code to be generated as
```js
const TableLists = styled(Script)({})
```
instead of 
```js
const TableLists = styled('Script')({})
```
because the `Script` is a react component that needs to be wrapped.


packages/teleport-test/src/standalone.ts
```diff
@@ -203,145 +202,6 @@ const run = async () => {
       })
       return `React - StyledComponents`
     })
-
```
ionutpasca, line 206: Is this supposed to be removed?

ionutpasca (approved):

packages/teleport-test/src/standalone.ts
JayaKrishnaNamburu: Yeah, we are not actively working on tailwind in generators anymore. So, it's just a lot of noise everytime we are testing the project developent.

PR #905, opened 2024-06-19
(feat): support named slots which helps in passing jsx to the components

### LINKED ISSUE #897: Support for UIDLElementNode as UIDLAttributeValue
The use case:

I want to be able to generate code that contains an element as an attribute of another element, one example using ReactJS would be using the component TextField from Material UI lib wich contains the prop "inputComponent" that can receive a component:

`<TextField
  InputProps={{
    inputComponent: MyInputComponent,
    inputProps: {
      component: SomeThirdPartyComponent,
    },
  }}
/>;`

This is quite simple to support to React, but I'm not sure about the other frameworks generators.
I have some ideas to do almost the same to Angular but I have no idea if this is possible to the others generators.

Can you share your thoughts about it? If you think it's viable I can work on it myself :)

JayaKrishnaNamburu:
Implement `named-slot` for `next` and `html` generators to pass jsx to components. 

```json
{
    "type": "element",
    "content": {
        "elementType": "component",
        "dependency": {
            "type": "local"
        },
        "referencedStyles": {},
        "abilities": {},
        "children": [],
        "attrs": {
            "heading": {
                "type": "element",
                "content": {
                  "elementType": "container",
                  "style": {
                    "padding": "20px"
                  },
                  "children": ["This is amazing"]
                }
              }
        },
        "semanticType": "Hero"
    }
}
```

Declaration of the prop is

```json
{
  "heading": {
      "defaultValue": "Magnificent things are very simple",
      "type": "element"
  }
}
```

This will help in creating jsx that can pass `jsx` as a prop to the component

```jsx
<Hero heading={<div>This is amazing</div>}></Hero>
```

```jsx
const Hero = () => {
  return <div className="hero-text">{props.heading}</div>
}
```


JayaKrishnaNamburu:
Related to #897

PR #917, opened 2024-07-31
(fix): css class name collisions between nodes

ionutpasca:


JayaKrishnaNamburu (approved):
All good, please fix the tests @ionutpasca

PR #922, opened 2024-08-08
(chore): improve github workflows for testing and publish

JayaKrishnaNamburu:
The issue with previous setup is, we are just building at the time of publish and not building the generated projects for cms. So, if just the packages pass the build they are released in publish workflow. And won't wait until the `cms` projects were actually built. This will make sure, publish runs only when the the cms projects were also built.


JayaKrishnaNamburu:
You can confirm them here, the tasks related to npm are skipped on all branches
https://github.com/teleporthq/teleport-code-generators/actions/runs/10298294631/job/28503222418?pr=922

PR #960, opened 2025-02-16
Support Dynamic values and Expression as operand of a Conditional Element

vladgrecu:
https://linear.app/teleporthq/issue/THQ-4151/[code-generators]-support-dynamic-values-and-expression-as-operand-of

### DIFF
--- a/packages/teleport-plugin-common/src/node-handlers/node-to-html/utils.ts
+++ b/packages/teleport-plugin-common/src/node-handlers/node-to-html/utils.ts
@@ -7,6 +7,8 @@ import {
   UIDLAttributeValue,
   UIDLEventHandlerStatement,
   UIDLElementNode,
+  UIDLExpressionValue,
+  UIDLDynamicReference,
 } from '@teleporthq/teleport-types'
 import { HTMLTemplateGenerationParams, HTMLTemplateSyntax } from './types'
 import { createHTMLNode } from '../../builders/hast-builders'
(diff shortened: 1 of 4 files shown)


packages/teleport-plugin-common/src/node-handlers/node-to-html/utils.ts
```diff
@@ -205,5 +207,17 @@ const stringifyConditionalExpression = (
     return `${identifier} ${operation} '${value}'`
   }
 
+  if (typeof value === 'number') {
+    return `${identifier} ${operation} ${value}`
+  }
+
+  if (value.type === 'dynamic') {
+    return `${identifier} ${operation} ${value.content.id}`
```
JayaKrishnaNamburu, line 215: For `dynamic`. If i remember correct we need to use `options.dynamicPrefix` to prefix the value with `props` by default or with a different name when needed.

vladgrecu: I'm actually thinking that we don't even need to treat this flow. In what code generation pipeline is it used? For non React frameworks?

JayaKrishnaNamburu: Even for reach framework too, in `content.id` we store only the `id` value. So, using `dynamicPrefix` we can get generated code like `props.id`

PR #962, opened 2025-03-20
fix/THQ-4188 Support prop drilling in HTML export

vladgrecu:
https://linear.app/teleporthq/issue/THQ-4188/[code-generators]-drilled-props-do-not-work-in-html-export


vladgrecu:
@JayaKrishnaNamburu think you can check it out? 
I remember you assessing this to be a bigger chunk of work but I found a way to do it pretty fast.
Any concerns you might have?

JayaKrishnaNamburu (approved):
The flattening for children seems to work as expected. I tested by passing a page prop to a component, that is passing to another child component here

[project.json](https://github.com/user-attachments/files/19432875/project.json)

Seems to work as expected and the tests are passing. Nice work and fix 🙌 
Screenshot 2025-03-24 at 9 02 11 PM

PR #968, opened 2025-04-02
fix/THQ-4188 Do not generate in HTML, the element, if attr value is false and default prop value is true

vladgrecu:
https://linear.app/teleporthq/issue/THQ-4188/[code-generators]-drilled-props-do-not-work-in-html-export

### DIFF
--- a/packages/teleport-plugin-html-base-component/src/node-handlers.ts
+++ b/packages/teleport-plugin-html-base-component/src/node-handlers.ts
@@ -162,16 +162,13 @@ export const generateHtmlSyntax: NodeToHTML<
           if (usedProp === undefined || usedProp.defaultValue === undefined) {
             return conditionalNodeComment
           }
-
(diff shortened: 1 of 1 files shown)


packages/teleport-plugin-html-base-component/src/node-handlers.ts
```diff
@@ -657,14 +659,34 @@ const generateComponentContent = async (
       }
     }
 
+    if (attribute?.type === 'expr') {
+      const [ctxId, ...refPath] = attribute.content.split('?.')
```
JayaKrishnaNamburu, line 663: little catch, if these expressions are added by end users like this `item.name` instead of `item?.name`. There might be a case it fails. But, i don' think thats the case right. These expressions are ususally auto generated by builder. So, It should be fine.

JayaKrishnaNamburu (approved):

PR #969, opened 2025-04-07
Conditional Rendering for project classes

vladgrecu:
https://linear.app/teleporthq/issue/THQ-4189/conditional-rendering-for-project-classes


packages/teleport-types/src/uidl.ts
```diff
@@ -758,6 +765,7 @@ export interface UIDLElementNodeProjectReferencedStyle {
   content: {
     mapType: 'project-referenced'
     referenceId: UIDLProjectReferencedStyleID
+    renderingConditions?: UIDLDynamicCondition
```
JayaKrishnaNamburu: This might be being picky, but we can call this `condition` and the `condition` inside of this as `expression`

PR #997, opened 2025-11-25
New data sources support

ionutpasca:


vladgrecu (approved):
Pe cand e gata gata....poate niste teste ar fi bune

PR #1071, opened 2026-09-09
feat: drop custom watcher in favour of turborepo

JayaKrishnaNamburu:
Not sure, if you still need this. But, i couldn't stop myself when i started using turbo again. I tried it multiple times on this before. 
https://github.com/teleporthq/teleport-code-generators/pull/661
https://github.com/teleporthq/teleport-code-generators/pull/730

Every single time, the approach fell short. Its because, turbo didn't have watch mode before. And here is a issue that's opened around the time i opened these above PR's.

https://github.com/vercel/turborepo/issues/986
https://github.com/teleporthq/teleport-code-generators/pull/661#issuecomment-1070984895

Screenshot 2026-09-09 at 5 20 15 PM

This can solve a couple of things.
- When the repo is started in watch mode. It always runs the `build` script. As we don't know, if the build was run before, and the `dist` inside the packages that already has its, is upto date or went state. So, we ended up running `build` everytime we started watch mode. `turbo` takes care of those things.
- We watch for the package, and build only the specific package. We don't do any upstream or the down stream package builds as its totally custom one. 
- The custom watcher can be dropped, and we don't need to wait `~1.5min`every time the repo need to be started.

https://github.com/user-attachments/assets/b9ade80e-2318-4ace-bdbc-39947cf4752f


JayaKrishnaNamburu:
Let me know, if you would like the change. I will cross-verify everything.

Zhiyilang074811:
Great move to turborepo! Cleaner builds, ready to merge.

